PR #666, opened 2022-12-16
feat: Add Toggle Group Component

### LINKED ISSUE #524: feature: Add ability to show closed, open, and all PRs on the calendar.
### Type of feature

Select the type of feature request, the lowercase should also be the PR prefix.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

### Current behavior

Similar to https://github.com/open-sauced/insights/issues/476, we should have an option to show, open/closed, and all PRs. This is a user flow I use a lot to explain the potential of the product. I need still use gs-api to do this. 

Screen Shot 2022-10-11 at 5 19 01 PM

### Suggested solution

This could be a sub-nav or tab, or switch. Leaving it to your interpretation @pixelsbyeryc 

### Code of Conduct

By submitting this issue, you agree to follow our [Code of Conduct](https://docs.opensauced.pizza/contributing/code-of-conduct/).

- [ ] I agree to follow this project's Code of Conduct

### Contributing Docs

If you plan on contributing code please read - [Contribution Guide](https://docs.opensauced.pizza/contributing/introduction-to-contributing/)

- [ ]  I agree to follow this project's Contribution Docs

Deadreyo:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

This PR is mainly to add a Toggle Group in Dashboard screen that can choose to only show data related to closed PRs, opened ones or all.

This feature needed a Toggle Group component, and there wasn't one so I've been working on making it first. I wonder if it should be separated into 2 PRs or just 1.

- [x] Toggle Component
- [ ] Filter PRs by state

- Toggle Component
         - An example is made in `nivo-scatter-chard.tsx` which will be replaced by the actual PR filtering.
         - Built on Radix UI toggle component (imported its module)
         - Styled using Tailwind
         - Added some documentation for it. Specifically documented the props of the component & and provided an example of its usage in the component's jsdocs description. I think this is a good step towards documenting our work, and could be great for everyone to slowly pick up. (Disregard the Draft comment, this is just a draft for my ideas and will be removed later on).
         - Implementation documentation is written below:
```md
This component is supposed to allow as many number of options as demanded

    * There was 2 ways to implement this:
        - Either by forcing other developers to use the ToggleGroupItem components inside the ToggleGroup component
            - i.e. by returning an error if the children are not ToggleGroupItem components
            - this allows for less flexibility
            - developers may forget that they need to use the ToggleGroupItem component, which is less intuitive
            - didnt think about how I would implement the logic control for it. Probably much harder

        - Or by allowing any type of component to be passed as children, but wrapping those components in a ToggleGroupItem component manually
            - this allows for more flexibility, and more control over the children
            - developers now wont need to remember that they need to use the ToggleGroupItem component and can use anything, which is more intuitive
            - this is the implementation I went with

    * This implementation depends on the input children being static, or dynamic but without re-ordering the children
        - as it uses the index of the children as keys, due to the possible absense of other unique identifiers
        - possibly, a key prop could be added to children, but I am not sure if you can enforce that prop to be present with TypeScript
        - if enforcing a prop for input children is possible, then we could also enforce a value prop, instead of using their indices as values.
```

~closes # 524~
part of #524 

A screenshot of the design in Campsite: 
image

A screenshot of Toggle Component: 
image

A screenshot of the toggle component that filters the PRs (final state): Not yet

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook


brandonroberts:
The implementation you chose is reasonable. I would also recommend the input children being dynamic, but not-reordered in the case we want to toggle certain selections off in the future

brandonroberts:
And 2 separate PRs would be preferred. One for adding the component to design system, and one for implementing the feature in the app

Deadreyo:
> The implementation you chose is reasonable. I would also recommend the input children being dynamic, but not-reordered in the case we want to toggle certain selections off in the future

Could you elaborate more on the meaning of "children being dynamic"?

brandonroberts:
I was referring to the child elements you pass into the component for display that you were talking about wrapping/not wrapping

Deadreyo:
> And 2 separate PRs would be preferred. One for adding the component to design system, and one for implementing the feature in the app

So I guess at this state this PR is ready for review?

Deadreyo:
@brandonroberts I've also thought about using uuid library to generate unique keys for each child. This will solve the "no re-ordering allowed" issue at the cost of having to re-render the whole children tree every time there's a change to the children (most likely wouldnt matter much though)

components/atoms/ToggleGroupItem/toggle-group-item.tsx
```diff
@@ -0,0 +1,26 @@
+import * as GroupToggle from "@radix-ui/react-toggle-group";
+import clsx from "clsx";
+
+interface ToggleGroupItemProps {
+    checked: boolean,
```
brandonroberts, line 5: nit: semi-colons instead of commas for interfaces

```suggestion
    checked: boolean;
```

components/atoms/ToggleGroup/toggle-group.tsx
```diff
@@ -0,0 +1,59 @@
...
+ * Warning: Do NOT re-order the children after the component has been rendered.
+ *
+ * @example
+ * <ToggleGroup>
+ *      <>Option 1</> // React Fragment. Will use our styling
+ *      <>Option <span>2</span></> // Can be a nested component
+ *      <div>Option 3</div> // Any type of component can be used
+ * </ToggleGroup>
+ */
+const ToggleGroup = ({
```
brandonroberts, line 25: This component needs an example added to the Storybook also. There are example stories are here https://github.com/open-sauced/insights/tree/beta/stories/atoms

brandonroberts (changes requested):

brandonroberts:
> @brandonroberts I've also thought about using uuid library to generate unique keys for each child. This will solve the "no re-ordering allowed" issue at the cost of having to re-render the whole children tree every time there's a change to the children (most likely wouldnt matter much though) what do you think?

I don't think we need to introduce uuids for that. If it becomes an issue, we'll revisit it.

components/atoms/ToggleGroup/toggle-group.tsx
```diff
@@ -0,0 +1,59 @@
...
+      handleChange && handleChange(value);
+    }
+  };
+
+  if(!children) return null;
+
+  return (
+    <GroupToggle.Root type="single" value={value} onValueChange={handleValueChange} className="bg-light-slate-6 rounded-lg p-0.25">
+      {
+        Array.isArray(children) ? children.map((child, index) => (
```
NsdHSO, line 45: It is not hard to read these branches? Hint `if(Array,isArray){....}  else{....}`

components/atoms/ToggleGroup/toggle-group.tsx
```diff
@@ -0,0 +1,59 @@
+import * as GroupToggle from "@radix-ui/react-toggle-group";
+import { useState } from "react";
+import ToggleGroupItem from "../ToggleGroupItem/toggle-group-item";
+
+interface ToggleGroupProps {
+    children: React.ReactNode,
+    /** Allows the user to deselect an option, resulting in none being selected. */
+    allowNone?: boolean,
```
NsdHSO, line 8: A descriptive name is much cleaner than many lines of comments.

Deadreyo: Got suggestions for names?
Those also aren't normal comments, they are JSDoc comments/description that will show up for each of those property when hovered on, whenever a developer tries to implement them anywhere

Deadreyo: image

components/atoms/ToggleGroup/toggle-group.tsx
Deadreyo: You can't make an if-statement inside a JSX block though

components/atoms/ToggleGroup/toggle-group.tsx
Deadreyo: Added ToggleGroup to Storybook. However ToggleGroupItem can't be added as a standalone component for test as it will give error if not put inside a ToggleGroup component, but it is included in ToggleGroup tests anyways

components/atoms/ToggleGroup/toggle-group.tsx
bdougie: I thin this is readable enough. No need to change. The real test for readability will be when it is live using real data.

bdougie (approved):
Looks good to me.

brandonroberts (approved):

components/atoms/ToggleGroup/toggle-group.tsx
NsdHSO: One possibility is to use IIFE.

PR #693, opened 2022-12-20
feat: repository selection & add to insights

### LINKED ISSUE #592: bug: Click to un-filter by repo
### Describe the bug

This was reported by @sw-yx. When clicking a repo it filters, but he mentions its unintuitive. Perhaps this easter needs more context. 

Screen Shot 2022-11-14 at 2 20 43 PM

### Suggested Solution
When a repo is filter, clicking that repo again should remove the filter. 

### Environment

Select the places where the issue can be reproduced (that you know of).

- [ ] Local
- [ ] Testing
- [ ] Development
- [ ] Production

### Code of Conduct

By submitting this issue, you agree to follow our [Code of Conduct](https://docs.opensauced.pizza/contributing/code-of-conduct/).

- [ ] I agree to follow this project's Code of Conduct

### Contributing Docs

If you plan on contributing code please read - [Contribution Guide](https://docs.opensauced.pizza/contributing/introduction-to-contributing/)

- [ ]  I agree to follow this project's Contribution Docs

### LINKED ISSUE #593: feature: Save to dashboard feature
image
### Type of feature

Select the type of feature request, the lowercase should also be the PR prefix.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

### Current behavior

When viewing the insights dashboard, a user should be able to click on a repo page to see more details.  Currently, we only show the repo tab, but it would be good to get custom page insights

### Suggested solution

We should nudge the user to create a custom dashboard by doing the following.

1. Sign in
2. Select a repo from the table or save to dashboard button

### Code of Conduct

By submitting this issue, you agree to follow our [Code of Conduct](https://docs.opensauced.pizza/contributing/code-of-conduct/).

- [ ] I agree to follow this project's Code of Conduct

### Contributing Docs

If you plan on contributing code please read - [Contribution Guide](https://docs.opensauced.pizza/contributing/introduction-to-contributing/)

- [ ]  I agree to follow this project's Contribution Docs

related https://github.com/open-sauced/insights/issues/592

Deadreyo:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

- Feature: ability to select 1 or more repositories from the Repository tab.
- Selected Repositories can be used as items for a new insights page for the user.
- Feature: clicking on an already filtered-by repo will remove the filter

- Unsigned users clicking on "Add to Insight" will be lead to login, then after logging in will be redirected to `insights/hub` with the selected repos.
> Implementation: Since logging in always leads to homepage and does a page refresh, we will use session storage to save that we want to redirect to a certain link and also save the selected repos, then in homepage we will check if there's a saved redirect link. If there is, we will redirect to it instead.

fixes #593
fixes #592 

image

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook


bdougie:
This is an excellent first pass. I have a few questions. 
1. Why is the button green?
2. This feature should redirect someone to log in at some point. We can't create a page without knowing the user. This should be available for users not logged in. My suggestion is when the user clicks the "Add to Inishgt Page."
3. What happens when you select more than 10 repos?

I also found a unrelated bug while QAing this https://github.com/open-sauced/insights/issues/704

Deadreyo:
> 1. Why is the button green?
that's the default color design of the pre-existing components in the project: checkbox & button. Thought it was intentional for consistency.

> 2. This feature should redirect someone to log in at some point. We can't create a page without knowing the user. This should be available for users not logged in. My suggestion is when the user clicks the "Add to Inishgt Page."
Yes sure.
> 3. What happens when you select more than 10 repos?
It works for any number of repos. Tested with 20 repos and it works. The implementation isn't affected by the number of repos
> I also found a unrelated bug while QAing this #704
I just tested it and it works on the live site.

bdougie:
> I also found a unrelated bug while QAing this https://github.com/open-sauced/insights/issues/704

> I just tested it and it works on the live site.

Confirmed this is only an issue on beta

Deadreyo:
> The click-to-filter feature was removed in this PR. Related - https://github.com/open-sauced/insights/issues/592

Well that's wierd. It's supposed to automatically merge that with this PR

bdougie:
> 2. This feature should redirect someone to log in at some point. We can't create a page without knowing the user. This should be available for users not logged in. My suggestion is when the user clicks the "Add to Inishgt Page."

For clarification. The feature should be useable to select repos, just redirecting the user when they try to navigate to the create page. There may be a small challenge in storing the saved repos after the redirect. 

i.e. I choose 3 repos and get redirected when I hit create to login in. After logging in, I should see the repos in the shopping cart on the create page.

Deadreyo:
I've noticed some janky/buggy behavior in some cases of selecting repos, so I'll add more fixes

Deadreyo:
> For clarification. The feature should be useable to select repos, just redirecting the user when they try to navigate to the create page. There may be a small challenge in storing the saved repos after the redirect.
> 
> i.e. I choose 3 repos and get redirected when I hit create to login in. After logging in, I should see the repos in the shopping cart on the create page.

So currently I made it so that it will ask to log in.
For the second part which is the redirecting to the original destination, I think that needs an upgrade to the sign in system itself to:
- Allow for custom redirects after signing in.
- Redirect to the same page the log in was requested from if it wasn't stated otherwise.

As this was also a clunky user experience I faced when using the app, which is that logging in will redirect to a different path and not return to the original path. Definitely will be easier to be done through the log in system itself

Deadreyo:
So I've had a look in UseSupabaseAuth hook, and I've found this
```
return {
    signIn: (data: SignInWithOAuthCredentials) => supabase.auth.signInWithOAuth({
      ...data,
      options: {
        redirectTo: process.env.NEXT_PUBLIC_BASE_URL ?? "/"
      }
    }),```
  The part
```options: {
        redirectTo: process.env.NEXT_PUBLIC_BASE_URL ?? "/"
      }```
is basically hard-coding the sign-in to redirect to the root path, even if a custom redirect was passed to it. Is there a reason why it is implemented that way?

**Comment by bdougie** [comment]
@brandonroberts can shed light on it. It came down to the authentication not playing nice with next.

**Review by bdougie — CHANGES_REQUESTED** [review]
1. Match the button color to orange
2. Prevent this feature from showing unless the user is logged in
3. Ensure this feature still works https://github.com/open-sauced/insights/issues/592

**Comment by brandonroberts** [comment]
@Deadreyo that may have been a side effect of the auth integration we had at the time. You can modify the method signature to take the additional options and apply them to the sign in redirect.

File: components/organisms/RepositoriesTable/repositories-table.tsx [hunk 1]
```diff
@@ -82,7 +87,7 @@ const RepositoriesTable = ({
 
       {isLoadedWithRepos &&
         listOfRepositories.map((item, index) => (
-          <RepoRow key={`${item.handle}/${item.name}/${index}`} topic={topic} repo={item} user={user} />
+          <RepoRow key={`${item.handle}/${item.name}/${index}`} topic={topic} repo={item} user={user} selected={selectedRepos.find((r) => r.id == item.id) != undefined} handleOnSelectRepo={handleOnSelectRepo} />
```
0-vortex, line 90: this is kind of mismatching the fact that the type is set to array but we are transporting undefined, why not force the variable to be an empty array instead of doing the filtering here?

Deadreyo: sorry, type of what exactly? which variable are you talking about?

Deadreyo:
> @Deadreyo that may have been a side effect of the auth integration we had at the time. You can modify the method signature to take the additional options and apply them to the sign in redirect.

I've tried playing around a lot with the `redirectTo` option but it always failed to work. Even changed the above to
```ts
signIn: (data: SignInWithOAuthCredentials) => supabase.auth.signInWithOAuth({
      ...data,
      options: {redirectTo: 'http://localhost:3000/hub/insights'}
    }),
```
and still leads to `/`

bdougie:
Would be able to use the api folder and manage this with a serverless function?

Deadreyo:
> Would be able to use the api folder and manage this with a serverless function?
> 
> https://www.netlify.com/blog/2021/12/11/serverless-functions-made-simple-just-add-files/

Hmm, not sure how that can help here. Signing in happens by going to a github page from the frontend so not sure if it's doable to change the signing in method with a serverless function (that's the usecase I thought of related to serverless functions)

Deadreyo:
The browser is also refreshed on logging in, so any state would be cleared. However, I do have an idea.

So since the behavior of logging is to refresh the page (which means all states are lost) and redirect to `/` url, I could make it so that before the sign in happens, I would save a "redirect to" url value to local storage, and in base `/` url, check if there's a "redirect to" value or not, and if yes lead to it. The same method can be used to save the selected repos to load after the redirect.

bdougie:
I'd say the only way to find out is by testing this on your branch. You can always revert if needed. Realisticall folks are not around today and tomorrow to review.

components/organisms/RepositoriesTable/repositories-table.tsx
0-vortex: repo can be an array of undefined repos

Deadreyo: if you mean the parameter "repo", that takes a single repo object and is always defined.

I think you mean the "selected" parameter. Well the line `selectedRepos.find((r) => r.id == item.id) != undefined` evaluates to a boolean, so there isn't any arrays being passed here.
Excuse me if that's not what you meant.

bdougie (approved):
I tested and it works. Some edge cases I tried as well
- select 2 repos, click to filter down to one, create insights button includes both.
- paginate while selecting
- logged out user is logged in after clicking the "create insights page"

Waiting on a second approval

bdougie:
The design includes an orange highlight when selected. Was there a blocker in implementing that @Deadreyo? Do you think we should implement that separately?

image

I plan on opening an issues for the filter feature.

Deadreyo:
> The design includes an orange highlight when selected. Was there a blocker in implementing that @Deadreyo? Do you think we should implement that separately?
> 
> image
> 
> I plan on opening an issue for the filter feature.

Changing the background color would be simple. Changing the borders tho will be more complicated and require separate logic for it as it would have to check if there are selections beside selections (and so cancel the borders between them) and such checks. Better to be done separately.

Deadreyo:
There's an edge case in which if you select all repositories (with the top checkbox) and then switch page, or increase display count, it would still be checked even when the displayed items are not selected (and not checked).

I'd prefer if this gets a separate issue as well to think about what the behavior in that case should be, and as its implementation could be complicated or more messy (from my estimation).

bdougie:
> There's an edge case in which if you select all repositories (with the top checkbox) and then switch page, or increase display count, it would still be checked even when the displayed items are not selected (and not checked).
> 
> 
> 
> I'd prefer if this gets a separate issue as well to think about what the behavior in that case should be, and as its implementation could be complicated or messy.
> 
> 
> 
> Edit: The behavior i have in mind: only if all repos currently displayed are selected, then the top checkbox will be selected. If some repos are selected but not all and top checkbox is selected, all unselected repos displayed will be added to selection.
> 
> 
> 
> Things to note: Selection may include repos not currently displayed, especially when changing pages.
> 
> This means the top checkbox could be checked if all displayed repos are selected, but there may be more repos selected as well. (as the top checkbox would give the meaning that only those displayed repos are selected, but in fact they may not be the only ones)

The design accounts for this with the "Filtered Selected Repos" button. That button should show currently selected repos. But agreed that should be addressed in a new issue opened after this merges.

components/organisms/RepositoriesTable/repositories-table.tsx
brandonroberts, line 90: I don't think the `selected` logic should be inline in the prop. Open up the `map` callback function, use a variable and pass it to the prop instead.

pages/index.tsx
```diff
@@ -31,7 +32,15 @@ const Home: WithPageLayout = () => {
       }
 
       if (currentUser?.data?.session && onboarded === true) {
-        redirect("/hub/insights");
+        const redirectUrl = sessionStorage.getItem("redirectUrl");
```
brandonroberts, line 35: Would prefer to use the `redirectTo` from the auth provider to go directly back to the referring page after login

brandonroberts (changes requested):

components/organisms/Repositories/repositories.tsx
```diff
@@ -37,6 +39,36 @@ const Repositories = ({ repositories }: RepositoriesProps): JSX.Element => {
...
+    }
+  };
+
+  const handleOnAddtoInsights = () => {
+    if(user) {
+      router.push({pathname: "/hub/insights/new", query: {selectedRepos: JSON.stringify(selectedRepos)}}, "/hub/insights/new" );
+    } else {
+      sessionStorage.setItem("redirectUrl", "/hub/insights/new");
+      sessionStorage.setItem("selectedRepos", JSON.stringify(selectedRepos));
+      signIn({ provider: "github" });
```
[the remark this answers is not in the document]
Deadreyo: `redirectTo` isn't working: https://github.com/open-sauced/insights/pull/693#issuecomment-1364583840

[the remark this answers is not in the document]
Deadreyo: Cheecking session storage for redirecting only happens for authed users. So if they cancel the auth, the session item will be set but won't be used, so won't be redirected.

pages/index.tsx
Deadreyo: Same conversation as the other `redirectTo` conversation: it doesn't work so this is a workaround.

components/organisms/RepositoriesTable/repositories-table.tsx
```diff
@@ -47,11 +47,14 @@ interface RepositoriesTableProps {
...
   error: boolean;
   repo?: string | string[] | undefined[];
+  selectedRepos: RepositoriesRows[];
+  handleOnSelectRepo: (repo: RepositoriesRows) => void;
 }
 
 export const classNames = {
   row: "hidden md:flex gap-4    items-center py-3 px-6 odd:bg-white even:bg-light-slate-2",
   cols: {
+    checkbox: "w-[38px]",
```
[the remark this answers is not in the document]
Deadreyo: There isn't a class for exactly 38px, there is 36px & 40px. 
image

pages/index.tsx
0-vortex: unfortunately, this won't work without the `redirectTo` doing the job in both instances :<

0-vortex (changes requested):
The `redirectTo` option is the only way we can successfully finish the supabase auth flow. While we do have a next/link leak in a couple of places, we are also not try catching the sign in process in all the places where we are logging in.

In order for this PR to work, it is likely we need to adjust the session cache using the auth middleware and successfully differentiate between server and client issues.

Part of the reason this doesn't work in local development can be found in discussions: https://github.com/supabase/supabase/discussions/9061 and https://github.com/supabase/supabase/discussions/2842

When using the `redirectTo` parameter, the proper `NEXT_PUBLIC_BASE_URL` environment variable needs to be used, which is hardcoded in `useSupabaseAuth.ts` because it cannot be changed until we check the sessions server side; this also means that the existing redirect login would happen in the middleware.

It's a tough change that will prove valuable in practice :<

Deadreyo:
Reverted the redirect commit to unblock this PR. Redirect should be handled in a separate issue

0-vortex:
- [ ] Unsigned users clicking on "Add to Insight" will be lead to login, then after logging in will be redirected to insights/hub with the selected repos.


I've investigated the issue of logging in random parts of the app and I don't think it is possible to consistently redirect to the second session managed system by just using `redirectTo`. Similarly to what @Deadreyo was describing as not working, is actually a more complex redirect system that can lose the origin header in multiple scenarios - this makes it impossible to redirect to something other than `/` without a middleware system accounting for server and client differences of where the user was in the onboarding process or other similar global state remembering where you were.

For this reason, I think we should abort the selection of repos before logging in, hint at the checkbox with a disabled state, and in case the user clicks to log in, would be redirected to the same page (from the middleware), where one could now select the repos to add to the insight page. This invalidates current checks for `user` and removes a lot of the boilerplate code to check for that in subsequent `onChange` functions 🍕

Deadreyo:
> I think we should abort the selection of repos before logging in, hint at the checkbox with a disabled state

Done.

0-vortex (approved):
Code-wise LGTM! 👍 

Testing the disabled checkbox now am wondering if some form of extra feedback is needed, like the transition from automatically logging in and being sent to the wrong page versus the checkbox not doing anything while logged, but also not hinting at the user having to log in. Not sure what the best approach is here, let's wait for the triage team to test this out in the UI as well ❤️ 🍕

brandonroberts (approved):
LGTM

brandonroberts:
Maybe some alt text on the checkboxes/button to "Connect to GitHub" would be enough on desktop at least

Deadreyo:
> Maybe some alt text on the checkboxes/button to "Connect to GitHub" would be enough on desktop at least

Added this 
image

Issue #677, opened 2022-12-17
Bug: overflow-x on the insights page


favourite1975:
### Describe the bug

![Screenshot from 2022-12-18 00-12-42](https://user-images.githubusercontent.com/72051159/208269528-468a868b-9ad8-4864-a046-de62a6f92371.png)
there is an overflow on the insight page for the large screen views

### Steps to reproduce

1. Go to https://insights.opensauced.pizza/react/repositories/filter/recent
2. drag left and right to see the error

### Affected services

opensauced.pizza

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

bdougie:
@favourite1975 thanks for opening this issue. Can you record a video or gif showing the error?

favourite1975:
[Screencast from 12-18-2022 01:40:47 AM.webm](https://user-images.githubusercontent.com/72051159/208271688-01ce39a3-017c-4a39-bcc4-0d7f6045751e.webm)

payn33:
.take

PR #709, opened 2022-12-22
fix: overflow x bug on dashboard

payn33:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description
- This pr fixes an overflow bug on the dashboard page by including an `overflow: hidden` property on the `filter.tsx` container

- Fixes: #677 

### Before 

https://user-images.githubusercontent.com/66219025/209103278-ba0a3c6a-e3e8-4a1e-b173-7c5a987ca8f0.mp4

### After
image

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/components/organisms/ToolList/nav.tsx
+++ b/components/organisms/ToolList/nav.tsx
@@ -22,7 +22,7 @@ const Nav: React.FC<NavProps> = ({ toolList, selectedTool = "dashboard", selecte
       aria-orientation="horizontal"
       aria-label="Browse the tools"
       tabIndex={0}
-      className="tool-list-nav flex w-[100vw] 2xl:w-full overflow-x-auto overflow-y-hidden gap-2 px-4 md:px-16 bg-light-slate-3 border-b pt-3"
+      className="tool-list-nav flex w-full overflow-x-auto overflow-y-hidden gap-2 px-4 md:px-16 bg-light-slate-3 border-b pt-3"
     >
       {toolList.map((tool, index) => (
         <div
--- a/layouts/filter.tsx
+++ b/layouts/filter.tsx
@@ -9,6 +9,7 @@ import FilterHeader from "components/molecules/FilterHeader/filter-header";
 import useNav from "lib/hooks/useNav";
 
 const FilterLayout = ({ children }: { children: React.ReactNode }) => {
+  
   const { toolList, selectedTool, filterName, selectedFilter, userOrg } = useNav();
 
   return (
@@ -28,7 +29,9 @@ const FilterLayout = ({ children }: { children: React.ReactNode }) => {
           />
         </div>
 
-        <main className="flex w-full flex-1 flex-col items-center px-3 md:px-16 py-8 bg-light-slate-2">{children}</main>
+        <main className="flex w-full flex-1 flex-col items-center px-3 md:px-16 py-8 bg-light-slate-2">
+          {children}
+        </main>
       </div>
       <Footer />
     </>
--- a/layouts/filter.tsx
+++ b/layouts/filter.tsx
@@ -16,7 +16,7 @@ const FilterLayout = ({ children }: { children: React.ReactNode }) => {
     <>
       <TopNav />
       <div className="page-container flex min-h-[calc(100vh-(54px+95px))] flex-col items-center">
-        <div className="info-container min-w-full min-h-[100px]">
+        <div className="info-container w-full min-h-[100px]">
           <Header>
             <FilterHeader />
           </Header>
--- a/layouts/filter.tsx
+++ b/layouts/filter.tsx
@@ -8,14 +8,13 @@ import FilterHeader from "components/molecules/FilterHeader/filter-header";
 
 import useNav from "lib/hooks/useNav";
 
-const FilterLayout = ({children}: {children: React.ReactNode}) => {
-
+const FilterLayout = ({ children }: { children: React.ReactNode }) => {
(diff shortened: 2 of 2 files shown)


layouts/filter.tsx
```diff
@@ -8,14 +8,13 @@ import FilterHeader from "components/molecules/FilterHeader/filter-header";
...
-const FilterLayout = ({children}: {children: React.ReactNode}) => {
-
+const FilterLayout = ({ children }: { children: React.ReactNode }) => {
   const { toolList, selectedTool, filterName, selectedFilter, userOrg } = useNav();
 
   return (
     <>
       <TopNav />
-      <div className="page-container flex min-h-[calc(100vh-(54px+95px))] flex-col items-center">
+      <div className="page-container flex min-h-[calc(100vh-(54px+95px))] flex-col items-center overflow-x-hidden">
```
0-vortex, line 17: Thanks for opening a PR for this, however hiding the overflow is not the correct way to solve this problem; we should figure out what the underlying element that breaks the 100% width range and assess whether this is a platform bug.

For reference, can you let us know your operating system and browser so we can replicate a testing environment? 🍕

0-vortex (changes requested):
Thanks for opening a PR, unfortunately we need some more information to be able to test this on our end! 🍕

layouts/filter.tsx
payn33: sure, i'm on windows 11 using a chrome browser, overflow appears around 1300px. following your advice I took a closer look and noticed that changing the tooltip nav.tsx parent elements width property from 100vw to 100% also fixes the overflow without having to hide it.

components/organisms/ToolList/nav.tsx
```diff
@@ -22,7 +22,7 @@ const Nav: React.FC<NavProps> = ({ toolList, selectedTool = "dashboard", selecte
       aria-orientation="horizontal"
       aria-label="Browse the tools"
       tabIndex={0}
-      className="tool-list-nav flex w-[100vw] 2xl:w-full overflow-x-auto overflow-y-hidden gap-2 px-4 md:px-16 bg-light-slate-3 border-b pt-3"
+      className="tool-list-nav flex w-full overflow-x-auto overflow-y-hidden gap-2 px-4 md:px-16 bg-light-slate-3 border-b pt-3"
```
payn33, line 25: not the right fix, removes the horizontal scroll from nav

layouts/filter.tsx
payn33: hi @0-vortex you can review the changes I made?

layouts/filter.tsx
```diff
@@ -8,8 +8,7 @@ import FilterHeader from "components/molecules/FilterHeader/filter-header";
 
 import useNav from "lib/hooks/useNav";
 
-const FilterLayout = ({children}: {children: React.ReactNode}) => {
-
+const FilterLayout = ({ children }: { children: React.ReactNode }) => {
```
brandonroberts, line 11: nit: revert unnecessary changes to this file

components/organisms/ToolList/nav.tsx
```diff
@@ -22,7 +22,7 @@ const Nav: React.FC<NavProps> = ({ toolList, selectedTool = "dashboard", selecte
       aria-orientation="horizontal"
       aria-label="Browse the tools"
       tabIndex={0}
-      className="tool-list-nav flex w-[100vw] 2xl:w-full overflow-x-auto overflow-y-hidden gap-2 px-4 md:px-16 bg-light-slate-3 border-b pt-3"
```
brandonroberts, line 25: @SunGoldTech thoughts on this change?

brandonroberts (changes requested):

components/organisms/ToolList/nav.tsx
0-vortex: I'm pro 👍

SunGoldTech: LGTM... great work @payn33 🍕

SunGoldTech (approved):

layouts/filter.tsx
payn33: this has been done.

layouts/filter.tsx
```diff
@@ -8,15 +8,15 @@ import FilterHeader from "components/molecules/FilterHeader/filter-header";
...
+  
   const { toolList, selectedTool, filterName, selectedFilter, userOrg } = useNav();
 
   return (
     <>
       <TopNav />
       <div className="page-container flex min-h-[calc(100vh-(54px+95px))] flex-col items-center">
-        <div className="info-container min-w-full min-h-[100px]">
+        <div className="info-container w-full min-h-[100px]">
```
payn33, line 19: the reason for this change is that a parent element with no fixed width will not allow the children elements with overflow auto or scroll properties to flow properly, causing the parent element to go over the screen.

brandonroberts (approved):

PR #725, opened 2022-12-23
feat: display organization lookup error if organization is not found.

### LINKED ISSUE #700: feature: show an error message when the provided organization does not exist on onboarding
### Type of feature

🍕 Feature

### Current behavior

At the moment, when a user provides a non-existent organization, there is no visual feedback to let the user know that something is not right.

onboarding-edit

### Suggested solution

Show a user friendly error message to let the user know that the organization does not exist.
https://github.com/open-sauced/insights/blob/70abbbaad7c4ba25075f21580d53a51e7b874ea3/pages/start.tsx#L160-L163

### Additional context

_No response_

### Code of Conduct

### Contributing Docs

shamimbinnur:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description
This PR adds the feature to display an error message if the entered organization is not found or is invalid while completing the second step of onboarding.

Fixes #700  

https://user-images.githubusercontent.com/48399373/209354402-a484e0e9-20eb-434c-a23b-7af20f8c95d2.mov

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -21,7 +21,6 @@ const Home: WithPageLayout = () => {
 
     async function getAccess() {
       const currentUser = await supabase.auth.getSession();
-      
       if (!currentUser?.data?.session) {
         redirect("/javascript");
       }
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -21,7 +21,7 @@ const Home: WithPageLayout = () => {
 
     async function getAccess() {
       const currentUser = await supabase.auth.getSession();
-
+      
       if (!currentUser?.data?.session) {
         redirect("/javascript");
       }
--- a/pages/start.tsx
+++ b/pages/start.tsx
@@ -110,16 +110,10 @@ interface LoginStep2Props {
   setRepoList: Function;
 }
 
-enum OrgLookupError {
-  Initial = 0,
-  Invalid = 3,
-  Error = 4
-}
-
 const LoginStep2: React.FC<LoginStep2Props> = ({ handleLoginStep, setRepoList }) => {
   const { providerToken } = useSupabaseAuth();
   const [orgName, setOrgName] = useState("");
-  const [addOrgError, setAddOrgError] = useState<OrgLookupError>(OrgLookupError.Initial);
+  const [addOrgErrorCode, setAddOrgErrorCode] = useState<number>(0);
 
   captureAnayltics("User Onboarding", "onboardingStep2", "visited");
 
@@ -160,34 +154,26 @@ const LoginStep2: React.FC<LoginStep2Props> = ({ handleLoginStep, setRepoList })
         });
 
         setRepoList(repoList);
-        setAddOrgError(OrgLookupError.Initial);
+        setAddOrgErrorCode(0);
       }
 
       // If valid, go to next step
       handleLoginStep();
     } catch (e: any) {
       // If invalid, display error
-      console.error(e);
-      
-      if ( e.code  == 404) {
-        setAddOrgError(OrgLookupError.Invalid);
-      } 
-      else {
-        setAddOrgError(OrgLookupError.Error);
-      } 
+      setAddOrgErrorCode(e.status);
     }
   };
 
-  const getOrgLookupError = (code: OrgLookupError) => {
-    if (code === OrgLookupError.Error) {
-      return <Text>There was error retrieving this organization.</Text>;
-    }
-
-    if (code === OrgLookupError.Invalid) {
-      return <Text>The organization entered is invalid.</Text>;
-    }
-
-    return <></>;
+  const getOrgLookupError = (errorCode: number) => {
+    switch (errorCode) {
+    case 0: 
+      return <></>;
+    case 404:
+      return <Text className="bg-light-red-4 p-1">The organization is invalid</Text>;
+    default:
+      return <Text className="bg-light-red-4 p-1">Something went wrong</Text>;
+    };
   };
 
   return (
@@ -238,7 +224,7 @@ const LoginStep2: React.FC<LoginStep2Props> = ({ handleLoginStep, setRepoList })
             />
           </div>
 
-          {getOrgLookupError(addOrgError)}
+          {getOrgLookupError(addOrgErrorCode)}
 
           <Button onClick={handleAddPAT} type="primary" className="w-full mt-3 md:mt-0 h-10">
             Continue
--- a/pages/start.tsx
+++ b/pages/start.tsx
@@ -31,6 +31,7 @@ import useSession from "lib/hooks/useSession";
 import { captureAnayltics } from "lib/utils/analytics";
 import { getAvatarLink } from "lib/utils/github";
 import useStore from "lib/store";
+import { Http2ServerResponse } from "http2";
 
 type handleLoginStep = () => void;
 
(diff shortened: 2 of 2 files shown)


pages/index.tsx
```diff
@@ -21,7 +21,7 @@ const Home: WithPageLayout = () => {
 
     async function getAccess() {
       const currentUser = await supabase.auth.getSession();
-
+
```
brandonroberts, line 24: nit: Revert whitespace change
```suggestion

```

brandonroberts (approved):
LGTM

pages/index.tsx
shamimbinnur: @brandonroberts done

pages/start.tsx
```diff
@@ -153,16 +160,36 @@ const LoginStep2: React.FC<LoginStep2Props> = ({ handleLoginStep, setRepoList })
         });
 
         setRepoList(repoList);
+        setAddOrgError(OrgLookupError.Initial);
       }
 
       // If valid, go to next step
       handleLoginStep();
-    } catch (e) {
+    } catch (e: any) {
```
0-vortex, line 168: please remove the `any` and console logging here, provide a user alert if the message is important to the user

pages/start.tsx
```diff
@@ -153,16 +160,36 @@ const LoginStep2: React.FC<LoginStep2Props> = ({ handleLoginStep, setRepoList })
...
+      if ( e.code  == 404) {
+        setAddOrgError(OrgLookupError.Invalid);
+      } 
+      else {
+        setAddOrgError(OrgLookupError.Error);
+      } 
     }
   };
 
+  const getOrgLookupError = (code: OrgLookupError) => {
```
0-vortex, line 181: this code split is not re-usable, the actual error message is not passed around; please apply the conditionals in the display view in a way that the user can read the error message and be alerted of the error.

pages/start.tsx
```diff
@@ -427,7 +457,7 @@ const Login: WithPageLayout = () => {
         </section>
         <section className="w-full lg:max-w-[50%] p-9 rounded-lg lg:rounded-r-lg bg-white">
           {currentLoginStep === 1 && <LoginStep1 handleLoginStep={handleLoginStep} user={user} />}
-          {currentLoginStep === 2 && <LoginStep2 handleLoginStep={handleLoginStep} setRepoList={setRepoList} />}
+          {currentLoginStep === 2 &&  <LoginStep2 handleLoginStep={handleLoginStep} setRepoList={setRepoList} />}
```
0-vortex, line 460: this is just adding a space

0-vortex (changes requested):
Please focus more on the end-user experience than the developer one; while the typescript code is getting better and better, the user has no way to describe the error they are getting and require additional hand-holding for reading the error (not everyone has the developer console even enabled)

We will not be able to creep up many issues and PRs on top of each other like we did in hot.opensauced, for example pushing PRs to beta with console.log in them for visual purposes, so please, only mark the PR as ready when it is working as expected in the deploy preview URL; with long-lived branches and deploy previews, anything purely visual can be drafted and left open.

pages/start.tsx
shamimbinnur: This is the same logic used for the 'Repo Look-up error' on the third step of the onboarding. So I just tried to keep the consistency of the code structure. And this actually re-useable and helps logically show the error message to the user based on the HTTP response code. For example only if the response code is 404 it will show the message "The organization entered is invalid." otherwise it will show "There was an error retrieving this organization. I think these two types of error and the corresponding error message are enough to get a user to understand what's gone wrong. Also as it is reusable we can easily add a single line of code to add another type of error message.

pages/start.tsx
shamimbinnur: Console logging was there, I was confused about whether it is intended or not.

If I remove 'any' it does not allow me to access the "e.code" since it was required to get the error code. The error message appears to the user based on the error code.

pages/start.tsx
0-vortex: both the org lookup and repo lookup can be improved; can you improve the part that is related to this pr?

pages/start.tsx
0-vortex: catching infers Error; when you are throwing objects that are more than Error, you should type it more strictly, not less IE any.

please type e to an actua error and ensure only that can be thrown 🍕

pages/start.tsx
bdougie: Let's start by making error messages a hue or red. These pills have a hue for the background as well.

brandonroberts:
@shamimbinnur what's left to get this ready to merge?

shamimbinnur:
Getting this error if I change from "any" to any other type

Screenshot 2023-01-10 at 10 10 10 PM

shamimbinnur:
https://user-images.githubusercontent.com/48399373/211603717-f7a59e7e-439a-405d-872d-32ffe18ff7d2.mov

brandonroberts:
The GitHub API does throw an error though if the org is invalid. We may need need to type the error as `unknown` in the `try` and cast it as `Error & { status: number }` to be able to access the status code.

0-vortex:
> The GitHub API does throw an error though if the org is invalid. We may need need to type the error as `unknown` in the `try` and cast it as `Error & { status: number }` to be able to access the status code.

I disagree with this, octokit.request is not typed return nor error because it's not supposed to throw; if that was the case, try cathing those blocks is fine, the way it is set up now it's encapsulating more logic than needed.

Whereas the `response` variable should be checked for data and error, the code itself is mapping the data over an array assuming it will work. This will not, and instead of throwing an API error it will throw a message saying that null is not an object, that is generic message, an empty object `{}` which can't be typed. 

We simply cannot use try/catch as a magic rubber band for writing incomplete code and in general, throwing errors should be avoided in client environments.

shamimbinnur:
Maybe now we should think about a different logic for this.

brandonroberts:
`ocktokit.request` does return a rejected promise with an error, so using the try/catch with the await is what we need https://www.npmjs.com/package/@octokit/request#request

From the docs

If an error occurs, the promise is rejected with an error object containing 3 keys to help with debugging:

- error.status The http response status code
- error.request The request options such as method, url and data
- error.response The http response object with url, headers, and data

shamimbinnur:
> `ocktokit.request` does return a rejected promise with an error, so using the try/catch with the await is what we need https://www.npmjs.com/package/@octokit/request#request
> 
> From the docs
> 
> If an error occurs, the promise is rejected with an error object containing 3 keys to help with debugging:
> 
> * error.status The http response status code
> * error.request The request options such as method, url and data
> * error.response The http response object with url, headers, and data

@brandonroberts going to try this.

bdougie:
The design of the error message is not blocking. I have opened an issue to explore error message design in the entire app. I'll leave a comment inline on the copy.

https://github.com/open-sauced/insights/issues/781

pages/start.tsx
```diff
@@ -153,16 +154,28 @@ const LoginStep2: React.FC<LoginStep2Props> = ({ handleLoginStep, setRepoList })
...
+      setAddOrgErrorCode(e.status);
     }
   };
 
+  const getOrgLookupError = (errorCode: number) => {
+    switch (errorCode) {
+    case 0: 
+      return <></>;
+    case 404:
+      return <Text className="bg-light-red-4 p-1">The organization is invalid</Text>;
```
bdougie, line 173:
```suggestion
      return <Text className="bg-light-red-4 p-1">That organization was not found.</Text>;
```

Not found is a better explanation than "invalid"

bdougie (changes requested):

Issue #688, opened 2022-12-19
Feature: Public/Private insights badge


bdougie:
### Type of feature

🍕 Feature

### Current behavior

There is no badge identifying if a dashboard is public or private

### Suggested solution

Add a badge that checks and displays if a hub page is public or private

https://www.figma.com/file/OpVX6WT7dmWqnwRuEvADMF/OpenSauced-%E2%80%94-Insights-(insights.opensauced.pizza)?node-id=5176%3A6423&t=jZIlELgoFGnOMY8Y-0

<img width="1440" alt="Insight Page - Repositories - Period v04@2x" src="https://user-images.githubusercontent.com/5713670/208544051-dc71f66b-6df9-4b86-a3d7-74819b6b1462.png">

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [ ] I agree to follow this project's Contribution Docs

payn33:
@bdougie can I work on this?

bdougie:
I went ahead and added you to the @open-sauced/triage team. If you are able to assist in triaging and fixing future bugs, I would love your assistance. 

For this feature, please ensure to leverage the design system and ask questions. I have assigned you to the issue.

payn33:
Alright i'll be sure to ask questions on anything I'm not clear on, thanks.

bdougie:
Can you also me a message in the discord, so I can coordinate your roles there?

payn33:
> Can you also me a message in the discord, so I can coordinate your roles there?

Hi @bdougie, i've opened a pr for this issue and sent a message on discord.

PR #735, opened 2022-12-24
feat: add public/private insights badge

payn33:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description
This pr adds the badge component to the application and design system.

- Fixes: #688 

### Before
image

### After (private)
image

### After (public)
image

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [x] 📕 storybook


components/molecules/InsightHeader/insight-header.tsx
```diff
@@ -22,9 +23,15 @@ const InsightHeader = ({ insight, repositories }: InsightHeaderProps): JSX.Eleme
...
       <div className="header-info flex flex-col grow justify-center p-2">
-        <Title level={1} className="!text-3xl font-semibold tracking-tight text-slate-900">
-          {insight && insight.name || "Insights"}
-        </Title>
+        <div className="flex items-baseline">
+          <Title level={1} className="!text-3xl font-semibold tracking-tight text-slate-900 mr-[1rem]">
+            {insight && insight.name || "Insights"}
+          </Title>
+          {insight &&   
+          <Text className="flex text-xl self-end !text-light-slate-11 border border-light-slate-6 rounded-[26px] bg-light-slate-3 px-2.5 py-1 h-fit leading-3.5">
```
OgDev-01, line 31: This looks good visually, but the implementation can be enhanced a bit... instead of styling the `Text` component to act as a wrapper, wrapping the text in a div would reduce the `classNames` used here.

```suggestion
      <div className="flex  px-2.5 py-1 border border-light-slate-6 bg-light-slate-3 rounded-[26px]">
            // Text component can go in here
      </div>
```

components/molecules/InsightHeader/insight-header.tsx
```diff
@@ -22,9 +23,15 @@ const InsightHeader = ({ insight, repositories }: InsightHeaderProps): JSX.Eleme
         <ContextThumbnail size={120} ContextThumbnailURL={""}></ContextThumbnail>
       </div>
       <div className="header-info flex flex-col grow justify-center p-2">
-        <Title level={1} className="!text-3xl font-semibold tracking-tight text-slate-900">
-          {insight && insight.name || "Insights"}
-        </Title>
+        <div className="flex items-baseline">
```
OgDev-01, line 26:
```suggestion
        <div className="flex items-baseline gap-4">
```

OgDev-01 (commented):
Looks good visually

components/molecules/InsightHeader/insight-header.tsx
OgDev-01: Using the `gap-4` class would handle the space between the Page Name and the Label, but not really sure 4 would match though... you can try out other sizes :) 😇

bdougie (changes requested):
@payn33 we leverage atomic design for this application. This requires small components and elements to be added as atoms to create molecules. Please see the [readme](https://github.com/open-sauced/insights#-atomic-design) for more context, but also refactor this badge into its own dynamic component in the atoms folder.

components/molecules/InsightHeader/insight-header.tsx
payn33: this was very helpful thanks, gap 4 was just the right property to use.

components/atoms/InsightBadge/insight-badge.tsx
```diff
@@ -0,0 +1,17 @@
+import Text from "components/atoms/Typography/text";;
+
+interface BadgeProps {
+    isPublic: boolean
+}
+
+const Badge = ({ isPublic }: BadgeProps): JSX.Element => {
+  return (
+    <div className="inline-flex border border-light-slate-6 rounded-[26px] bg-light-slate-3 px-2.5 py-1">
+      <span className="text-sm text-light-slate-11 leading-none">
```
payn33, line 10: replacing the `Text` component with a span because it adds unnecessary line height that breaks the badge
cc: @bdougie @0-vortex

bdougie: Thanks for moving this to its component and add the story

bdougie (approved):

components/atoms/InsightBadge/insight-badge.tsx
payn33: sure, i went through the readme like you suggested 😅

components/atoms/InsightBadge/insight-badge.tsx
```diff
@@ -0,0 +1,17 @@
+import Text from "components/atoms/Typography/text";;
+
+interface BadgeProps {
+    isPublic: boolean
+}
+
+const Badge = ({ isPublic }: BadgeProps): JSX.Element => {
+  return (
+    <div className="inline-flex border border-light-slate-6 rounded-[26px] bg-light-slate-3 px-2.5 py-1">
```
NsdHSO, line 9: Please avoid using rounded-[26px] this is an anti-pattern in **open-sauced**. In tailwind, the builtin  max class is `.rounded-3xl` and in px is `16px`

bdougie: I agree with this. This eventually doesn't scale.

bdougie (changes requested):

components/atoms/InsightBadge/insight-badge.tsx
payn33: Since tailwind does not have a builtin class that matches the value given on the design (26px) is it acceptable to create a global style and apply there or use the built in max class which is also visually acceptable?

bdougie: Yes, that is fine. A global class makes it scale if we need to change in multiple places. 

please name it `rounded-badge`

0-vortex: disagree with global classes, use rem if this isn't design-system related (it's not), the problem is not using responsive values.

components/molecules/InsightHeader/insight-header.tsx
```diff
@@ -22,10 +23,13 @@ const InsightHeader = ({ insight, repositories }: InsightHeaderProps): JSX.Eleme
         <ContextThumbnail size={120} ContextThumbnailURL={""}></ContextThumbnail>
       </div>
       <div className="header-info flex flex-col grow justify-center p-2">
-        <Title level={1} className="!text-3xl font-semibold tracking-tight text-slate-900">
-          {insight && insight.name || "Insights"}
-        </Title>
-        <div className="flex mt-4 items-center relative  gap-2">
+        <div className="flex items-end gap-4">
+          <Title level={1} className="!text-3xl font-semibold tracking-tight text-slate-900">
```
0-vortex, line 27: I see the `text-3xl` class was marked as `!important` - is this still required with the new changes?

0-vortex (changes requested):
The title component force sets the font-size to xl using important and breaks the scaling down of components, that is also the reason why creating a span and setting that to `text-sm` works, but it's a hack. The title component should set the correct font-size in the parrent and not use any `!important` classes or rescale the text in a child component - doing this will also make it possible to use 1.5 size gaps correctly, as they will be 3 times bigger whet the baseline text is the intended `xl` size and not the current `sm` size.

Hard no on creating any custom or global classes for this. Limited/wrong code should not be reused in creating a new separate badge component, however, in practice, this translates to exactly 2 lines of code changed size-wise (different text and rem classes tho).

bdougie:
@payn33, checking in to see if you have questions. A few more things needed to get this mergeable.

components/molecules/InsightHeader/insight-header.tsx
payn33: No, it's no longer required.

bdougie (approved):
LGTM

0-vortex (approved):

PR #771, opened 2023-01-10
feat: add custom profile banner for verified OpenSauced user 

### LINKED ISSUE #713: feature: Set up a hook or check to validate if the user view has an opensauced account
This should be a simple implementation to check if the user profile is an OpenSauced user. 

The visual representation is should be the header card changed to the rainbow.

e508b85a-fb22-4d9c-a529-e0e2aec626ba (1)

https://www.figma.com/file/OpVX6WT7dmWqnwRuEvADMF/OpenSauced-%E2%80%94-Insights-(insights.opensauced.pizza)?node-id=5204%3A6170&t=dNJd6vghHX3zbPBV-0

OgDev-01:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description
This PR:
- creates a hook to validate if a user has an open-sauced account
- implement header changes when a user is open-sauced user

Fixes #713 

### Open sauced user
image

### Non Open sauced user
image

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook


next-types.d.ts
```diff
@@ -107,3 +107,17 @@ interface DbUserInsightRepo {
...
+  readonly id: number;
+  readonly open_issues: number;
+  readonly is_private: boolean;
+  readonly is_open_sauced_member: boolean;
+  readonly created_at: string;
+  readonly updated_at: string;
+  readonly login: string;
+  readonly is_onboarded: boolean;
+  readonly is_waitlisted: boolean;
+  readonly role: number;
```
0-vortex, line 121: please only add the fields you need here, we don't want to maintain parity between fields that are not used!

lib/hooks/useValidateUser.ts
```diff
@@ -0,0 +1,17 @@
+import useSWR from "swr";
+
+interface UserResponse{
+  data: DbUser
+}
+const useValidateUser = (username:string)=>{
```
0-vortex, line 6: The naming and the logic for this are off: the hook itself should test the `user.is_open_sauced_member` once it is fetched, while we don't have the data this is trying to do 2 things in one.

Please separate the data fetching patterns from checking the returned data itself: https://swr.vercel.app/docs/data-fetching

lib/hooks/useValidateUser.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  const {data, error} = useSWR<UserResponse, Error>(`user/${username}`);
```
brandonroberts, line 8: This is supposed to be hitting the API. Example here

https://github.com/open-sauced/insights/blob/beta/lib/hooks/useUserInsights.ts#L13

next-types.d.ts
```diff
@@ -107,3 +107,17 @@ interface DbUserInsightRepo {
   readonly repo_id: number;
   readonly created_at: string;
 }
+
+interface DbUser{
```
brandonroberts, line 111: nit: formatting

lib/hooks/useValidateUser.ts
```diff
@@ -0,0 +1,17 @@
+import useSWR from "swr";
```
brandonroberts, line 1: nit: formatting

brandonroberts (changes requested):

0-vortex (changes requested):
We only have 5 MB of cache to use with localstorage, we should use that to manage the information that is displayed on the initial load. Using the SWR pattern further, we have to configure it to the context that we are using it in.

The function to check member status can be ```const isOpenSaucedMember = (user: DbUser) => {
    return !!user.is_open_sauced_member;
  }```

The hook is needed in order to be used around multiple components, however this hook would not really be useful if it would check varying usernames. However, that hook behaviour is only useful for things that can be cached, since this feature looks like it is distinguishing users on the profile page, all the ticket should do would be to change the existing data fetching pattern from gitsense to the public api, or both in tandem! 🍕

bdougie:
@OgDev-01 can you implement this on the user profile to display a different header is the user is verified in this PR?

image

0-vortex:
SWR data fetching docs can be reviewed at https://swr.vercel.app/docs/data-fetching

OgDev-01:
> @OgDev-01 can you implement this on the user profile to display a different header is the user is verified in this PR?

Sure!, this can be done after this PR is merged.

components/molecules/ContributorProfileHeader/contributor-profile-header.tsx
```diff
@@ -1,15 +1,38 @@
...
+  isOpenSaucedUser: boolean;
 }
-const ContributorProfileHeader = ({ avatarUrl }: ContributorProfileHeaderProps) => {
+const ContributorProfileHeader = ({ avatarUrl, githubName, isOpenSaucedUser }: ContributorProfileHeaderProps) => {
   return (
-    <div className="w-full flex items-end px-6 md:px-10 lg:px-16 bg-[#DFE3E6] h-[216px]">
-      <div className="translate-y-[70px]">
-        <Avatar hasBorder avatarURL={avatarUrl} size={184} isCircle />
+    <div className="w-full relative  bg-light-slate-6 h-[216px]">
+      <div className={`relative w-full h-full ${!isOpenSaucedUser && "hidden"}`}>
```
0-vortex, line 16: please add conditional rendering here instead of hiding it with CSS

components/molecules/ContributorProfileHeader/contributor-profile-header.tsx
```diff
@@ -1,15 +1,38 @@
...
+        <Image priority alt="user profile cover image" layout="fill" objectFit="cover" src={RainbowBg} />
+      </div>
+
+      <div className="w-full absolute -top-6 px-6 md:px-12 lg:px-16 flex flex-row items-end justify-between py-6">
+        <div className="translate-y-[75px]">
+          <Avatar className="" hasBorder avatarURL={avatarUrl} size={184} isCircle />
+        </div>
+        <div className={`flex gap-3 flex-col md:flex-row items-center ${!isOpenSaucedUser && "hidden"}`}>
```
0-vortex, line 24: please add conditional rendering here instead of hiding it with CSS

components/organisms/ContributorProfilePage/contributor-profile-page.tsx
```diff
@@ -65,87 +75,93 @@ const ContributorProfilePage = ({
...
-                    <Text className="!text-lg md:!text-xl lg:!text-2xl !text-black !leading-none">{openPrs} PRs</Text>
-                  </div>
-                ) : (
-                  <div className="flex justify-center items-end mt-1"> - </div>
-                )}
+      {loading && <>Loading...</>}
+      {error && <>An error has occured...</>}
+      {isLoaded && (
+        <>
+          <ContributorProfileHeader isOpenSaucedUser={!!user} githubName={githubName} avatarUrl={githubAvatar} />
```
0-vortex, line 87: this should be `isOpenSaucedUser={!!user.is_open_sauced_member}`?

0-vortex (changes requested):
Couple of small changes! 👍

bdougie:
> @brandonroberts and @bdougie, I need more in a more detailed explanation of the data needed for verification to complete the implementation of this PR... I have made provisions for the verified user header on the design system

We need to verify if the user has "connected their GitHub" account to opensauced. 

While looking at this, curious why the check takes so long and if can it be improved. I am waiting anywhere from 2-5 seconds. 

Screen Shot 2023-01-12 at 2 19 59 PM

brandonroberts:
It's waiting on the contributor data and the user check to load. The contributor data takes significantly longer to load. I don't think we need to block the profile on that and just let it fill in when the contributor data is loaded.

components/organisms/ContributorProfilePage/contributor-profile-page.tsx
OgDev-01: The check for `user.is_open_sauced_member` is meant for member's of the org... Per our discussions in the engineering sync yerterday,  once  a user is returned from the api call, they are opensauced users, so we don't have to do additional checks for `is_open_sauced_member` 🍕

OgDev-01:
>While looking at this, curious why the check takes so long and if can it be improved. I am waiting anywhere from 2-5 seconds.

I can make a skeleton loader for areas that need to wait until API data is ready instead of the `Loading...` as the contributor endpoint takes significant time to load. 

This can be moved to a separate PR for now, so this can be merged

components/organisms/ContributorProfilePage/contributor-profile-page.tsx
0-vortex: That is not correct, we have users which have `is_open_sauced_member` false. We will always have users like that because we cannot control who owns a repo.

0-vortex (approved):
One more change 🙏 💯

pages/user/[username].tsx
```diff
@@ -8,25 +8,37 @@ import { ContributorsProfileType } from "components/molecules/ContributorHoverCa
...
+  const { data: user, isLoading: userLoading, isError: userError } = useFetchUser(contributorLogin);
 
+  const isLoading = contributorLoading || userLoading;
+  const isError = contributorError;
   const repoList = useRepoList(contributor[0]?.recent_repo_list || "");
   const contributorLanguageList = (contributor[0]?.langs || "").split(",");
   const profile: ContributorsProfileType = {
     githubAvatar: `https://www.github.com/${contributorLogin}.png?size=300`,
-    githubName: contributorLogin,
+    githubName: contributor[0]?.host_login,
```
Comment (brandonroberts) [inline pages/user/[username].tsx:31]: Keep this as-is.

```suggestion
    githubName: contributorLogin
```

pages/user/[username].tsx
```diff
@@ -8,25 +8,37 @@ import { ContributorsProfileType } from "components/molecules/ContributorHoverCa
...
 
-  const { data: contributor, isLoading: contributorLoading, isError } = useSingleContributor(contributorLogin);
+  const {
+    data: contributor,
+    isLoading: contributorLoading,
+    isError: contributorError
+  } = useSingleContributor(contributorLogin);
+  const { data: user, isLoading: userLoading, isError: userError } = useFetchUser(contributorLogin);
 
+  const isLoading = contributorLoading || userLoading;
```
Comment (brandonroberts) [inline pages/user/[username].tsx:25]: Just use `isLoading` from line 23. We don't want to block the UI on the contributor call to finish.

```suggestion
  const isLoading = contributorLoading || userLoading;
```

components/organisms/ContributorProfilePage/contributor-profile-page.tsx
brandonroberts: We should change this prop to `isVerified={!!user}` or `isConnected={!!user}` so its not so specific. Verified/connected just means they've connected their GitHub account

brandonroberts (changes requested):

brandonroberts (approved):
LGTM

Issue #724, opened 2022-12-23
Bug: repositories table not responsive on screen size between 1020px and 1285px


payn33:
### Describe the bug

- There's no provision for the responsiveness of the repositories table between the 1020 and 1285px 

### above 1285px 
![image](https://user-images.githubusercontent.com/66219025/209322462-200c6bcf-d4b9-42e3-ba82-ee3f2390da68.png)

### above 1020px but below 1285px
![image](https://user-images.githubusercontent.com/66219025/209322564-54312407-78db-467a-9307-438e79a421bc.png)

### Steps to reproduce

1.  Go to https://insights.opensauced.pizza/javascript/repositories/filter/recent
2.  Adjust the screen size anywhere between 1020px and 1285px

### Affected services

opensauced.pizza

### Platforms

Desktop

### Browsers

Chrome

### Environment

Production

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

PR #773, opened 2023-01-12
fix: repositories table responsiveness bug

payn33:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Fixes: #724 

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/components/organisms/Repositories/repositories.tsx
+++ b/components/organisms/Repositories/repositories.tsx
@@ -95,105 +95,107 @@ const Repositories = ({ repositories }: RepositoriesProps): JSX.Element => {
         setRangeFilter={store.updateRange}
         title="Repositories"
       />
-      <div className="flex flex-col rounded-lg overflow-hidden border">
-        <div className="flex md:hidden justify-between  py-4 px-6 bg-light-slate-3 gap-2">
-          <div className="flex-1">
-            <TableTitle text="Repository"></TableTitle>
+      <div className="flex flex-col rounded-lg overflow-x-auto border w-full">
+        <div className="min-w-[1150px]">
+          <div className="flex md:hidden justify-between  py-4 px-6 bg-light-slate-3 gap-2">
+            <div className="flex-1">
+              <TableTitle text="Repository"></TableTitle>
+            </div>
+            <div className="flex-1">
+              <TableTitle text="Pr Overview"></TableTitle>
+            </div>
           </div>
-          <div className="flex-1">
-            <TableTitle text="Pr Overview"></TableTitle>
+          <div className="hidden md:flex py-4 px-6 bg-light-slate-3 gap-2">
+            <div className={clsx(classNames.cols.checkbox)} >
+              <Checkbox label="" onChange={handleOnSelectAllChecked} disabled={!user} title={!user? "Connect to GitHub" : ""} className={`checked:[&>*]:!bg-orange-500 ${ user? "[&>*]:!border-orange-500 [&>*]:hover:!bg-orange-600": "[&>*]:!border-light-slate-8"}`}/>
+            </div>
+            <div className={clsx(classNames.cols.repository)}>
+              <TableTitle text="Repository"></TableTitle>
+            </div>
+            <div className={clsx(classNames.cols.activity)}>
+              <TableTitle text="Activity"></TableTitle>
+            </div>
+            <div className={clsx(classNames.cols.prOverview)}>
+              <TableTitle text="PR Overview"></TableTitle>
+            </div>
+            <div className={clsx(classNames.cols.prVelocity)}>
+              <TableTitle text="PR Velocity"></TableTitle>
+            </div>
+            <div className={clsx(classNames.cols.spam)}>
+              <TableTitle text="SPAM"></TableTitle>
+            </div>
+            <div className={clsx(classNames.cols.contributors, "hidden lg:flex")}>
+              <TableTitle text="Contributors"></TableTitle>
+            </div>
+            <div className={clsx(classNames.cols.last30days, "hidden lg:flex")}>
+              <TableTitle text="Last 30 Days"></TableTitle>
+            </div>
           </div>
-        </div>
-        <div className="hidden md:flex py-4 px-6 bg-light-slate-3 gap-2">
-          <div className={clsx(classNames.cols.checkbox)} >
-            <Checkbox label="" onChange={handleOnSelectAllChecked} disabled={!user} title={!user? "Connect to GitHub" : ""} className={`checked:[&>*]:!bg-orange-500 ${ user? "[&>*]:!border-orange-500 [&>*]:hover:!bg-orange-600": "[&>*]:!border-light-slate-8"}`}/>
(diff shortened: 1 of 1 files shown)


components/organisms/Repositories/repositories.tsx
```diff
@@ -95,105 +95,107 @@ const Repositories = ({ repositories }: RepositoriesProps): JSX.Element => {
...
-          <div className={clsx(classNames.cols.prVelocity)}>
-            <TableTitle text="PR Velocity"></TableTitle>
-          </div>
-          <div className={clsx(classNames.cols.spam)}>
-            <TableTitle text="SPAM"></TableTitle>
-          </div>
-          <div className={clsx(classNames.cols.contributors, "hidden lg:flex")}>
-            <TableTitle text="Contributors"></TableTitle>
+      <div className="flex flex-col rounded-lg overflow-x-auto border w-full">
+        <div className="lg:min-w-[1150px]">
```
payn33, line 99: this is the minimum value for a fully responsive table and a breakpoint was used to prevent the min-width from creating a horizontal scroll on medium and small screen table layouts

brandonroberts:
@payn33 will you post before and after screenshots so we can see what the applied changes look like

components/organisms/Repositories/repositories.tsx
0-vortex: agree with the class, not sure wrapping all the content in a wrapper is needed tho since we want to override `w-full` with `lg:min-w-[1150px]`

0-vortex (changes requested):
Thanks for a great PR, the wrapping extra `div` class should be used on the parent, that should make this good for merging! 👍

payn33:
> Thanks for a great PR, the wrapping extra `div` class should be used on the parent, that should make this good for merging! 👍

That would be the table parent right? because adding the wrapper to the main div would include the table header in the horizontal scroll  @0-vortex 
image

OgDev-01 (approved):
LGTM 🍕 , just some example screenshots of before and after, and we should be good on this

components/organisms/Repositories/repositories.tsx
OgDev-01: with the implementation, a div was used to prevent the table header from scrolling with the table when a horizontal scroll is happening on the table, and I think that is a good approach on this one 👍

0-vortex:
> > Thanks for a great PR, the wrapping extra `div` class should be used on the parent, that should make this good for merging! 👍
> 
> That would be the table parent right? because adding the wrapper to the main div would include the table header in the horizontal scroll @0-vortex image

If the new wrapper limits only `lg` sizes, we have just moved the problem from "looking ugly on small screen" to "looking ugly on big screens". If the parent spans to full width, the child component cannot be a fixed width if that makes sense. Removing the child component or adjusting the `w-full` parent, the idea is to not break visual cohesion solving a bug (fixing a bug with a new bug) ❤️

NsdHSO:
IMHO, you should catch this breakpoint `1020px and 1285px` and slice the columns, what column to be shown, and what column would be hidden.

something like that ( just pseudo code )    

    state = { windowWith: window.innerWidth } 
    handleResize = (event) => { 
        this.setState({ windowWith: window.innerWidth }) 
    }
  
    componentDidMount() { 
        window.addEventListener(‘resize’, this.handleResize) 
    } 

Good PR.

0-vortex:
> IMHO, you should catch this breakpoint `1020px and 1285px` and slice the columns, what column to be shown, and what column would be hidden.
> 
> 
> 
> something like that ( just pseudo code )    
> 
> 
> 
>     state = { windowWith: window.innerWidth } 
> 
>     handleResize = (event) => { 
> 
>         this.setState({ windowWith: window.innerWidth }) 
> 
>     }
> 
>   
> 
>     componentDidMount() { 
> 
>         window.addEventListener(‘resize’, this.handleResize) 
> 
>     } 
> 
> 
> 
> Good PR.  👍 
> 
> 

That is not the expected or intended behaviour since we are using tailwind: https://tailwindcss.com/docs/responsive-design

Please refrain from suggesting anti-patterns!!!

NsdHSO:
> > IMHO, you should catch this breakpoint `1020px and 1285px` and slice the columns, what column to be shown, and what column would be hidden.
> > something like that ( just pseudo code )
> > ```
> > state = { windowWith: window.innerWidth } 
> > 
> > handleResize = (event) => { 
> > 
> >     this.setState({ windowWith: window.innerWidth }) 
> > 
> > }
> > 
> > 
> > 
> > componentDidMount() { 
> > 
> >     window.addEventListener(‘resize’, this.handleResize) 
> > 
> > } 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Good PR.  👍
> 
> That is not the expected or intended behaviour since we are using tailwind: https://tailwindcss.com/docs/responsive-design
> 
> Please refrain from suggesting anti-patterns!!!

With this approach, we have something to implement 
image


Tailwind, is just for CSS, to make more easy to work with CSS and avoid some diff approaches to about CSS. But in IMO in this PR he must manage the columns.  To hide or show... But if you are not ok with this approach, please forget this intervention.  

But for a moment it is not possible, I understand.

payn33:
> > > Thanks for a great PR, the wrapping extra `div` class should be used on the parent, that should make this good for merging! 👍
> > 
> > 
> > That would be the table parent right? because adding the wrapper to the main div would include the table header in the horizontal scroll @0-vortex image
> 
> If the new wrapper limits only `lg` sizes, we have just moved the problem from "looking ugly on small screen" to "looking ugly on big screens". If the parent spans to full width, the child component cannot be a fixed width if that makes sense. Removing the child component or adjusting the `w-full` parent, the idea is to not break visual cohesion solving a bug (fixing a bug with a new bug) ❤️

Hi @0-vortex this is the current behaviour
1150px

In the proposed fix the child component is only a fixed width on `lg`  sizes which causes an overflow due to the parent element having the `overflow-x-auto` property at the min responsive width `1150px`  excluding the table header from the horizontal scroll 

https://user-images.githubusercontent.com/66219025/212764413-530c306b-bbf7-481e-86c5-7ad527c2c61e.mp4

Removing the child component and adding  the `overflow-x-auto` property to the `w-full` parent  includes the table header, is this what you mean by `not breaking visual cohesion` ?

https://user-images.githubusercontent.com/66219025/212765254-b543140a-9aa1-4c81-91ad-268f29f83324.mp4

OgDev-01 (approved):

PR #838, opened 2023-02-07
fix: enable user setting inputs

### LINKED ISSUE #837: feature: Make All Settings Editable
Noticing that the user cannot update bio, twitter, etc. Also email & timezone does not stay updated when saved. 

All fields should be able to be updated.

Screen Shot 2023-02-07 at 7 12 32 AM

Deadreyo:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

Enables user setting input fields & fix related issues.

closes #837 
## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/components/atoms/TextInput/text-input.tsx
+++ b/components/atoms/TextInput/text-input.tsx
@@ -14,6 +14,7 @@ interface TextInputProps {
   classNames?: string;
   errorMsg?: string;
   value?: string;
+  defaultValue?: string;
   onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
 }
 
@@ -29,6 +30,7 @@ const TextInput = ({
   autoFocus,
   borderless = false,
   value,
+  defaultValue,
   onChange,
   errorMsg = ""
 }: TextInputProps) => {
@@ -57,6 +59,7 @@ const TextInput = ({
             placeholder={placeholder || ""}
             onChange={onChange}
             value={value}
+            defaultValue={defaultValue}
             className={`flex-1 focus:outline-none  ${classNames} ${
               disabled && "bg-light-slate-3 cursor-not-allowed  text-light-slate-9"
             }`}
--- a/components/organisms/UserSettingsPage/user-settings-page.tsx
+++ b/components/organisms/UserSettingsPage/user-settings-page.tsx
@@ -59,7 +59,7 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
       });
       setSelectedInterest(insightsUser?.interests?.split(","));
       setDisplayLocalTime(insightsUser?.display_local_time);
-      setTimezone(insightsUser?.timezone);
+      setTimezone(insightsUser?.timezone || "");
     }
     fetchAuthSession();
   }, [user, insightsUser]);
@@ -147,8 +147,7 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
                 rows={4}
                 placeholder="Tell us about yourself."
                 className="bg-light-slate-4 rounded-lg px-3 py-2 disabled:cursor-not-allowed "
-                readOnly
-                value={
+                defaultValue={
                   userInfo?.bio ||
                   "I am an open source developer with a passion for music and video games. I strive to improve the open source community and am always looking for new ways to contribute."
                 }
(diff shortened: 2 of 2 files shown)


components/organisms/UserSettingsPage/user-settings-page.tsx
```diff
@@ -158,28 +157,24 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
...
               label="URL"
-              disabled
             />
             <TextInput
               classNames="bg-light-slate-4 text-light-slate-11"
               placeholder="@saucedopen"
               label="Twitter Username"
-              disabled
-              value={`@${(userInfo && userInfo.twitter_username) || "saucedopen"}`}
+              defaultValue={`@${(userInfo && userInfo.twitter_username) || "saucedopen"}`}
```
brandonroberts, line 165: Why the switch from `value` to `defaultValue`? We'll also need API updates to enable these additional fields to be updated

Deadreyo: Setting `value` blocks the user from changing the value of these fields. 
`defaultValue` only sets the starting value of the field and allows the user to change it.

brandonroberts: Yes, but its not connected to the form, so it won't get captured with the form submission. We need to either

- Create a `useState` for each field to track the changes. Pass the state as `value` and update it when the input changes.
- Create a `useReducer` to track the overall state and update it for each field in the form. Then use the value from the state to submit with the form.

I would recommend the second option

Deadreyo: We could also use `useRef` to get the value of each input at form submission.

brandonroberts: I like that option also

Deadreyo:
Awaiting api updates then will continue on this.

brandonroberts:
@Deadreyo go ahead and continue with this feature. I'll post here when the API is ready

brandonroberts:
@Deadreyo the API endpoint is ready https://beta.api.opensauced.pizza/#/Authentication%20service/updateProfileForUser

Deadreyo:
> @Deadreyo the API endpoint is ready https://beta.api.opensauced.pizza/#/Authentication%20service/updateProfileForUser

Are all fields required?

components/atoms/TextInput/text-input.tsx
```diff
@@ -14,6 +14,10 @@ interface TextInputProps {
   classNames?: string;
   errorMsg?: string;
   value?: string;
+  defaultValue?: string;
+  required?: boolean;
+  innerRef?: any;
```
brandonroberts, line 19: Is there no generic that can be used here? `fieldRef` would be a better name also.

components/organisms/UserSettingsPage/user-settings-page.tsx
```diff
@@ -93,14 +113,26 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
     }
   };
 
-  const handleUpdateProfile = async () => {
+  const handleUpdateProfile = async (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
```
brandonroberts, line 117: Instead of using refs for all the fields, why not just get them from the `e.target.form`?

components/organisms/UserSettingsPage/user-settings-page.tsx
```diff
@@ -33,24 +33,44 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
...
+    // if (user) setEmail(user.email);
+    fetchAuthSession();
+  }, [user]);
 
-    if (user) setEmail(user.email);
-    if (insightsUser) {
+  // insightsUsers uses SWR which keeps triggering re-rendering and overwriting the user input. Disabling this for now
+  let firstInsightsLoad = false;
+  useEffect(() => {
+    if (insightsUser && !firstInsightsLoad) {
```
brandonroberts, line 73: I'd prefer this to be fixed in the hook if possible to not keep refreshing the data.

lib/hooks/update-user.ts
```diff
@@ -1,7 +1,11 @@
 import { supabase } from "lib/utils/supabase";
 
+export interface updateUserPayload {
```
brandonroberts, line 3: nit: formatting

Fields should be on separate lines.

brandonroberts (changes requested):

components/atoms/TextInput/text-input.tsx
brandonroberts: A generic is a type that you can specify, similar to `React.useRef<HTMLFormElement>` where the `T` in `React.useRef<T>()` is the generic you provide to give additional context to the shape of what you're providing.

components/organisms/UserSettingsPage/user-settings-page.tsx
Deadreyo: Turned out this is actually related to user object being re-assigned frequently from `useSupabaseAuth` hook. Should be addressed in a new issue.

Deadreyo: Or I can remove `user` from the useEffect dependency array if that won't cause side problems

brandonroberts (approved):
LGTM

OgDev-01 (approved):

PR #894, opened 2023-02-21
v1.30.0-beta.8 -> production

brandonroberts:


0-vortex (approved):
Image names should replace wildcards with dashes `-` for pages not involving 404 pages! 🍕 👍

PR #966, opened 2023-03-07
feat: add official logos for pre-defined topics.

### LINKED ISSUE #928: Feature: Pre-defined topics list should have custom logos
### Type of feature

🍕 Feature

### Current behavior

We treat Hacktoberfest special because it was the first topic, but since we have a pre-defined list of topics, we should provide the official logos for the languages. ML and Ai can leverage the pre-selected emojis

Screen Shot 2023-03-01 at 1 38 55 AM

### Suggested solution

Use better logic to update the logos similar to how we do for Hacktoberfest.

Screen Shot 2023-03-01 at 1 39 22 AM

### Additional context

_No response_

### Code of Conduct

### Contributing Docs

- [ ] I agree to follow this project's Contribution Docs

shamimbinnur:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description
This PR adds official logos for pre-defined topics.
Removed the existing logic for hacktoberfest. Created a function `getTopicThumbnail` which takes a topic name as a parameter and returns its official logo for the topic thumbnail.

Fixes #928 

Screenshot 2023-03-07 at 8 41 39 PM

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook


components/molecules/FilterHeader/filter-header.tsx
```diff
@@ -40,15 +36,14 @@ const HeaderFilter = () => {
...
       </div>
       <div className="header-info md:truncate flex flex-col grow justify-center p-2">
         <Title level={1} className="!text-3xl font-semibold tracking-tight text-slate-900">
-          {isHacktoberfest ? "Hacktoberfest 2022" : topicNameFormatting(filterName as string)}
+          {topicNameFormatting(filterName as string)}
         </Title>
         <Text className="mt-1 !text-base   text-slate-500">
-          Insights on GitHub repositories{" "}
-          {isHacktoberfest ? "opted into the largest open source hackathon." : `using the ${filterName} topic.`}
+          {`Insights on GitHub repositories using the ${filterName} topic.`}
```
brandonroberts, line 46:
```suggestion
          {`Insights on GitHub repositories using the ${topicNameFormatting(filterName as string)} topic.`}
```

lib/utils/getTopicThumbnail.ts
```diff
@@ -0,0 +1,73 @@
...
+import cpp from "../../img/topicThumbnails/cpp.svg";
+import php from "../../img/topicThumbnails/php.svg";
+import c from "../../img/topicThumbnails/c.svg";
+import ruby from "../../img/topicThumbnails/ruby.svg";
+import ai from "../../img/topicThumbnails/ai.svg";
+import ml from "../../img/topicThumbnails/ml.svg";
+import react from "../../img/topicThumbnails/react.svg";
+import contextThumbnailImage from "../../img/open-sourced-with-bg-icon.png";
+
+const topicThumbnails = [
```
brandonroberts, line 15: Make this a dictionary instead so lookups are faster

```js
{
  [topic]: thumbnail
}
```

brandonroberts (changes requested):
The images are of various sizes. We should have a consistent size for all the images, so the UI doesn't shift around

components/molecules/FilterHeader/filter-header.tsx
bdougie: If consistent images are needed, we can get @getaheaddev to do that in figma

shamimbinnur: Since these are SVG, can we achieve it by providing width and height in the SVG file? Also, [ContextThumbnailProps](https://github.com/open-sauced/insights/blob/beta/components/atoms/ContextThumbnail/context-thumbnail.tsx) takes `size` as its optional props, so we can also set a boundary for the size.

@bdougie @brandonroberts

lib/utils/getTopicThumbnail.ts
```diff
@@ -0,0 +1,36 @@
...
+import cpp from "../../img/topicThumbnails/cpp.svg";
+import php from "../../img/topicThumbnails/php.svg";
+import c from "../../img/topicThumbnails/c.svg";
+import ruby from "../../img/topicThumbnails/ruby.svg";
+import ai from "../../img/topicThumbnails/ai.svg";
+import ml from "../../img/topicThumbnails/ml.svg";
+import react from "../../img/topicThumbnails/react.svg";
+import contextThumbnailImage from "../../img/open-sourced-with-bg-icon.png";
+
+const topicThumbnails = {
```
Deadreyo, line 15: This should be synchronized with the array of topics returned in `utils/getTopicOptions.ts`. If we add another pre-defined topic, it should give an error if we forgot to add a corresponding image & vice versa.

Deadreyo (changes requested):
Just some connection between types needed.

lib/utils/getTopicThumbnail.ts
shamimbinnur: @Deadreyo  Yes, we should somehow synchronize this with the topic, and interest. But that's should be done in a different PR.
But at this moment this is not gonna throw any error if we add any predefined topic and do not add a corresponding image. Instead, it will show the Open Suaced logo.

Deadreyo: IMO should be done in this, since you're already at it. Not signifcant to require a separate PR.

Deadreyo: To clarify it more for you, since in `getTopicOptions` we have an array of the topics, and in your map you have the same topics as keys, you can look up the TS syntax that will generate a type out of the array that you can pass to the map.

brandonroberts:
@shamimbinnur I deleted `getTopicOptions` so there's only this one now 

https://github.com/open-sauced/insights/blob/beta/lib/utils/getInterestOptions.ts

lib/utils/getTopicThumbnail.ts
shamimbinnur: @Deadreyo  IMHO, what you're saying may not be feasible. While it's possible to generate a type from a one-dimensional array, this type cannot be used for an object. The elements of an array are string elements, whereas the elements of the object are acting like key-value pairs. Therefore, using the type of an array element for an object may not be appropriate.

We can get type out an array but that will just specify if the type of value is a string, number, or something. But here we do not need to know whether the interest array's elements are strings or numbers or not.

We need to know which elements are there on the interest list! The type of array is not gonna tell us that.

The interface might help here but maybe we need to have the same object structure for both `interests` and `thumbnails`

Deadreyo: It's actually possible:

```
const interests = ["javascript", "python", "java", "typescript", "rust", "csharp", "cpp", "php", "c", "ruby" , "ai", "ml", "react"] as const;

let map: Record<typeof interests[number], string> = {
  javascript: "javascript",
  ai: "ai",
  python: "python",
  java: "java",
  typescript: "typescript",
  rust: "rust",
  csharp: "csharp",
  cpp: "cpp",
  php: "php",
  c: "c",
  ruby: "ruby",
  react: "react",
  ml: "ml",
};```

Comment (Deadreyo) [reply lib/utils/getTopicThumbnail.ts:15]: You can create the type `typeof interests[number]` in `getInterestOptions.ts` and export it to use it outside. Needless to mention, you'll also need to move interests array out of the function to the global scope to export it.

Comment (Deadreyo) [reply lib/utils/getTopicThumbnail.ts:15]: https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type For more info

Comment (shamimbinnur) [reply lib/utils/getTopicThumbnail.ts:15]: @Deadreyo it works! Thanks!!

File: components/molecules/FilterHeader/filter-header.tsx [hunk 4]
```diff
@@ -40,20 +37,19 @@ const HeaderFilter = () => {
...
         <Text className="mt-1 !text-base   text-slate-500">
-          Insights on GitHub repositories{" "}
-          {isHacktoberfest ? "opted into the largest open source hackathon." : `using the ${filterName} topic.`}
+          {`Insights on GitHub repositories using the ${topicNameFormatting(filterName as string)} topic.`}
         </Text>
         <div className="flex mt-4 items-center gap-2">
           <FilterCardSelect
             selected={filterName as string}
-            options={topicOptions}
+            options={topicOptions as unknown as []}
```
Deadreyo, line 52: `topicOptions as unknown as []`
You're forced to do this as `topicOptions` now returns a readonly string[], while `options` asks only for a normal string[]. `options` doesn't change in the string array though, so instead of doing this type-casting, go to the definition of the prop `options` and convert its type to `readonly string[]`.

Deadreyo (changes requested):
One last thing for me

components/molecules/FilterHeader/filter-header.tsx
shamimbinnur: @Deadreyo  The second approach seems sustainable. After doing some quick research I got to know that

when we use as unknown as string[], we are telling TypeScript to trust that the readonly array is actually a string[] array at runtime, even though TypeScript cannot verify this. This can lead to runtime errors if the actual type of the array is not string[].

So instead of using  `as unknown as string[]` I used the array spreading technique here:

`const interests = ["javascript", "python", "java", "typescript", "csharp", "cpp", "php", "c", "ruby" , "ai", "ml", "react"] as const;`

`export const getInterestOptions = () => {
  return [...interests];
};`

Deadreyo: That's pretty good, nice move!

Deadreyo (approved):
LGTM

brandonroberts (approved):

PR #969, opened 2023-03-07
v1.33.0-beta.10 -> production

brandonroberts:


components/atoms/LanguagePill/LanguagePill.tsx
```diff
@@ -1,16 +1,24 @@
...
 import Image, { StaticImageData } from "next/image";
 
-import JavascriptIcon from "img/icons/☕️.svg";
+import JavascriptIcon from "img/icons/👾.svg";
 import ReactIcon from "/img/icons/⚛️.svg";
 import PythonIcon from "/img/icons/🐍.svg";
 import AIIcon from "/img/icons/🤖.svg";
 import MLIcon from "/img/icons/🧠.svg";
 import RustIcon from "/img/icons/🦀.svg";
+import PhpIcon from "/img/icons/📄.svg";
```
0-vortex, line 10: think we need an issue here to remove the icons that are also emojis, it's clearing a lot of technical debt on incremental improvements to this page

components/molecules/ContributorHighlight/contributor-highlight-card.tsx
```diff
@@ -89,8 +89,8 @@ const ContributorHighlightCard = ({ title, desc, prLink, user, id }: Contributor
         );
         setLoading(false);
         if (res) {
-          ToastTrigger({ message: "Highlights Updated Successfully", type: "success" });          
-          mutate(`users/${user}/highlights`);          
+          ToastTrigger({ message: "Highlights Updated Successfully", type: "success" });
+          mutate(`users/${user}/highlights`);
           setOpen(false);
           setTimeout(() => {
             document.body.setAttribute("style", "pointer-events:auto !important");
```
0-vortex, line 96: IMHO this is not part of this release but should be looked at maybe in a tooling issue, programatic styles are "meh" at best; even if we have to use the same method to apply the properties, we should work around tailwind classes rather than setting manual styles, applying the same pattern in multiple ui blocking components will result in chaos 😅

components/molecules/ContributorProfileInfo/contributor-profile-info.tsx
```diff
@@ -70,7 +70,7 @@ const ContributorProfileInfo = ({
               About
             </Title>
             <p className={clsx("text-light-slate-11 text-sm", !bio && "!text-light-slate-8")}>
-              {bio || (githubName+" has connected their GitHub but has not added a bio.")}
+              {bio || githubName + " has connected their GitHub but has not added a bio."}
```
0-vortex, line 73: string math could use a tooling improvement issue too

components/molecules/HighlightInput/highlight-input-form.tsx
```diff
@@ -153,7 +153,7 @@ const HighlightInputForm = (): JSX.Element => {
       {pullrequestLink && isDivFocused && <GhOpenGraphImg githubLink={pullrequestLink} />}
 
       {isDivFocused && (
-        <Button loading={loading} disabled={!bodyText || !validCharLimit()} className="ml-auto" variant="primary">
+        <Button loading={loading} disabled={!bodyText || !validCharLimit()} className="ml-auto " variant="primary">
```
0-vortex, line 156: stray spacing and some other eslint issues could be solved by applying some of the auto-format rules from the backends (we are ultimately using typescript the same way outside of react), don't remember if there was an issue for this 😅

lib/utils/topic-name-formatting.ts
```diff
@@ -7,6 +7,10 @@ const topicNameFormatting = (name: string) => {
   if(name.toLowerCase() == "javascript") return "JavaScript";
   if(name.toLowerCase() == "ai") return "AI";
   if(name.toLowerCase() == "ml") return "Machine Learning";
+  if(name.toLowerCase() == "cpp") return "CPP (C++)";
+  if(name.toLowerCase() == "csharp") return "CSharp (C#)";
+  if(name.toLowerCase() == "php") return "PHP";
+  if(name.toLowerCase() == "typescript") return "TypeScript";
```
0-vortex, line 13: unpopular opinion but these could use a switch, actually a good use case for it

0-vortex (approved):
LGTM! 👍 

Left comments only for opening some tooling improvement / developer experience / good first issues if you think that's worth, or just pass them to triage? 🍕

PR #1009, opened 2023-03-15
refactor: add a check for proper URL in avatarURL prop

### LINKED ISSUE #732: Bug: AvatarProps cannot be static data
### Describe the bug

The `AvatarProps` in `Avatar` and `CustomAvatar` component cannot be static since we are sending that data to cloudinary. This makes the component render at all times, since the second one is string, even for potentially broken images.

### Steps to reproduce

visit https://insights.opensauced.pizza/ and check the console for broken images

### Affected services

insights.opensauced.pizza

### Platforms

_No response_

### Browsers

_No response_

### Environment

Production, Development, Testing

### Additional context

_No response_

### Code of Conduct

### Contributing Docs

Deadreyo:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

- Adds a check for the input of avatarURL in `<Avatar>`. If the string isn't a url, it will be counted as if no input was given.
- Also refactors some small pieces of the component.

closes #732 
## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/components/atoms/Avatar/avatar.tsx
+++ b/components/atoms/Avatar/avatar.tsx
@@ -64,7 +64,7 @@ const CustomAvatar = ({
           alt={alt ? alt : "Avatar"}
           width={size as number}
           height={size as number}
-          src={avatarURL ? avatarURL : AvatarImage}
+          src={avatarURL}
         />
       ) : (
         <div
@@ -100,7 +100,7 @@ const DefaultAvatar = ({
           className={`${className ? className : ""} object-cover`}
           alt={alt ? alt : "Avatar"}
           fill={true}
-          src={avatarURL ? avatarURL : AvatarImage}
+          src={avatarURL}
         />
       ) : (
         <div
--- a/components/atoms/Avatar/avatar.tsx
+++ b/components/atoms/Avatar/avatar.tsx
@@ -8,7 +8,7 @@ interface AvatarProps {
   avatarURL?: string | StaticImageData;
   initials?: string;
   alt?: string;
-  size?: "sm" | "base" | "lg" | number;
+  size: "sm" | "base" | "lg" | number;
   hasBorder?: boolean;
   isCircle?: boolean;
   isCached?: boolean;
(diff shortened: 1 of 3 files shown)


bdougie (approved):

components/atoms/Avatar/avatar.tsx
```diff
@@ -20,19 +20,26 @@ const avatarLoader = () => {
 };
 
 const Avatar = (props: AvatarProps): JSX.Element => {
-  const imageSource =
+  if(typeof props.avatarURL === "string") new URL(props.avatarURL);
```
0-vortex, line 23: IMHO this is not doing anything?

components/atoms/Avatar/avatar.tsx
```diff
@@ -100,7 +107,7 @@ const DefaultAvatar = ({
           className={`${className ? className : ""} object-cover`}
           alt={alt ? alt : "Avatar"}
           fill={true}
-          src={avatarURL ? avatarURL : AvatarImage}
+          src={avatarURL}
```
0-vortex, line 110: Is `AvatarImage` redundant now?

components/atoms/Avatar/avatar.tsx
Deadreyo: `new URL(...)` will throw an error if the string inside isn't a proper url. 
image

components/atoms/Avatar/avatar.tsx
Deadreyo: Pretty much. It was redundant before the changes as well, since `avatarURL` would always be valid in that line.

Deadreyo: It was an image of hacktoberfest as well, so we don't really have a use for that anymore

components/atoms/Avatar/avatar.tsx
brandonroberts: I don't think we should thrown runtime errors that aren't caught and handled

OgDev-01: I agree with ted on this, the check is not leveraged anywhere in the component

```
try {
    // Checks if the avatarURL is a proper URL. If not, it will throw an error.
    if (typeof imageSource === "string") new URL(imageSource);
  } catch (error) {
    console.error(error);
    imageSource = undefined;
  }
```
The above line does the work IMHO

Deadreyo: Oh, I understand now, yeah that's definitely a mistake

brandonroberts (approved):

PR #1030, opened 2023-03-19
feat: add `ProfileLanguageChart` component to design system and user profile page

### LINKED ISSUE #531: bug: Contributor languages should be relative to the amount of code (in that language) they've written in last 30 days
## Summary

This is most likely to do with how the API is sending us data...

Our current implementation of the Contributor Languages seems a bit off to me. If we compare it to other services, like the one TED uses on his README, or the UI GitHub uses for repositories, ours seems off.

| @0-vortex's Profile | From Insights Repo | Our implementation |
| --- | --- | --- |
| Screen Shot 2022-10-12 at 3 57 19 PM | Screen Shot 2022-10-12 at 3 57 31 PM | Screen Shot 2022-10-12 at 3 56 00 PM |
| Shows percentages on each language relative by what the user has written in X amount of time | Shows the percentage of languages relative to all of the files available in the repo | Shows top 5 languages equally distributed |

I'm not sure why this is happening, but it seems to be due to how the API is giving us the data.

I would expect something like:
- User has edited 32 javascript files, 10 markdown files, and 5 html files in all of his PRs in the last 30 days.
  - `javascript`: 68%
  - `markdown`: 22%
  - `html`: 10%
- User has edited 20 javascript files, 15 typescript files, 5 html files, 5 css files, 1 C# files, 1 markdown file, in all of his PRs in the last 30 days.
  - `javascript`: 42%
  - `typescript`: 31%
  - `html`: 10%
  - `css`: 10%
  - `other`: 7%

OgDev-01:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description
This PR creates a new component for the user's language used chart

Closes #531 

image

image

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [x] 📕 storybook


0-vortex (changes requested):
The PR is assuming that the backend can send more than 5 languages when technically we have to make this value static. When doing so, we will have more "options" available from the point in time we calculate and flatten the values, for example before counting lines of code, we could count prs and commits that appear with that specific language; depending on the logic we use to determine this, all of the processing logic should happen on the backend, else we are biasing the api design by forcing the frontend feature to request a fast api and severely bias it rather than redesign both.

Should the frontend component calculate things like this all the time? Definitely no. Similarly, the backend api should have the number of days or number of languages you want retrieved back - at the moment, and the way github does it, there is a "languages" array in every repo, we'd have to iterate over all the user repos and factor that in, when in real life that is not accurate enough either.

When that information is computed, it should be stored in a value in the users table that the API can then get "statically" and be displayed on the user profile in the frontend. To we deal with this caching problem at all in the frontend? Or do all pages just hit the API all the time?

OgDev-01:
> The PR is assuming that the backend can send more than 5 languages when technically we have to make this value static. When doing so, we will have more "options" available from the point in time we calculate and flatten the values, for example before counting lines of code, we could count prs and commits that appear with that specific language; depending on the logic we use to determine this, all of the processing logic should happen on the backend, else we are biasing the api design by forcing the frontend feature to request a fast api and severely bias it rather than redesign both.
> 
> Should the frontend component calculate things like this all the time? Definitely no. Similarly, the backend api should have the number of days or number of languages you want retrieved back - at the moment, and the way github does it, there is a "languages" array in every repo, we'd have to iterate over all the user repos and factor that in, when in real life that is not accurate enough either.
> 
> When that information is computed, it should be stored in a value in the users table that the API can then get "statically" and be displayed on the user profile in the frontend. To we deal with this caching problem at all in the frontend? Or do all pages just hit the API all the time?

OgDev-01:
> To we deal with this caching problem at all in the frontend? Or do all pages just hit the API all the time?

The requests are being cached until the data's are changed

0-vortex:
> > To we deal with this caching problem at all in the frontend? Or do all pages just hit the API all the time?
> 
> The requests are being cached until the data's are changed

Don't see any caching happening in the user page:  https://insights.opensauced.pizza/user/0-vortex/contributions / https://github.com/open-sauced/insights/tree/beta/pages/user/%5Busername%5D - as far as I understand any user visiting this page will trigger fetching the data from the backend and repainting the UI

components/molecules/ProfileLanguageChart/profile-language-chart.tsx
```diff
@@ -0,0 +1,72 @@
...
+  useEffect(() => {
+    const totalSumOfFirstFivePercentage = sortedLangArray
+      .slice(0, 5)
+      .map((lang) => lang.percentageUsed)
+      .reduce((prev: number, next: number) => prev + next); // need some help fixing this type error, used any to bypass 🙏
+    setPercentage(totalSumOfFirstFivePercentage);
+  }, [percentage, sortedLangArray]);
+
+  return (
+    <div className="flex flex-col gap-1 min-w-[120px]">
```
NsdHSO, line 25: Could you not use anti pattern? Here you have all tailwind classes and their specs

OgDev-01: Thanks

components/molecules/ProfileLanguageChart/profile-language-chart.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+import colors from "../../../lib/utils/color.json";
+
+const languageToColor: AllSimpleColors = colors as AllSimpleColors;
+const NOTSUPPORTED = "#64748B";
+interface ProfileLanguageChartProps {
+  languageList: LanguageObject[];
+}
+const ProfileLanguageChart = ({ languageList }: ProfileLanguageChartProps) => {
+  const sortedLangArray = languageList.sort((a, b) => b.percentageUsed - a.percentageUsed);
```
brandonroberts, line 13: Create a copy of the original array before mutating it with `sort`

```suggestion
  const sortedLangArray = languageList.slice().sort((a, b) => b.percentageUsed - a.percentageUsed);
```

components/molecules/ProfileLanguageChart/profile-language-chart.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const [percentage, setPercentage] = useState<any>(0);
+
+  useEffect(() => {
+    const totalSumOfFirstFivePercentage = sortedLangArray
+      .slice(0, 5)
+      .map((lang) => lang.percentageUsed)
+      .reduce((prev: number, next: number) => prev + next); // need some help fixing this type error, used any to bypass 🙏
```
brandonroberts, line 20: Is this note still needed?

brandonroberts (changes requested):

components/molecules/ProfileLanguageChart/profile-language-chart.tsx
brandonroberts: Yes, but is the note about the error still needed

OgDev-01: Ooh... sorry, didn’t notice that still exist 😁

0-vortex (approved):
LGTM! 👍 

Will need minor tweaks once we actually have this information in the API 😅

NsdHSO:
✅ Very nice, LGTM

PR #1048, opened 2023-03-23
v1.37.0-beta.7 -> production

brandonroberts:


0-vortex (approved):
IMHO code refactors like story improvements, tooling, and enhancement rewrites should be released independently from features.

Feature-wise, there are some unneeded `!important` rules and some type/interface mismatch. Multiple paradigms in terms of writing code, functional programming, and more static assets are being introduced compared to the last release. 

Petition to stop adding in things that we don't need, that we can't solve at a release type merge - we are close to some of these issues being A+ level effort for any single person, meaning it is becoming more and more difficult for any given person to fix things like for example the CSS `!important` rules (they keep being added even for components that don't need them).

Focusing on visual feature development, the important classes make development unpredictable, while having multiple patterns to choose from in terms of how we are writing those components can be huge time sinks. If not addressed, those patterns, should at the very least, not become more complex with every release as they have a negative effect on average developer velocity! 😅

PR #1067, opened 2023-03-30
fix: adding regex pattern to validate twitter username input

### LINKED ISSUE #1033: Bug: Updating profile on Insights page results in an error
### Describe the bug

Updating your user settings, e.g. bio, LinkedIn url etc results in an error.

### Steps to reproduce

1. Navigate to your user settings, https://insights.opensauced.pizza/user/settings
2. Add some info you want to update to your profile, e.g. bio, LinkedIn url etc.
3. Click the Update Profile button
4. A toast message appears saying "An error occured!!!"

CleanShot 2023-03-20 at 22 46 06

Additional context:
The preflight request appears to run, but then the actual request gives a 404

CleanShot 2023-03-20 at 22 52 37

Here is the payload response:

```
{"statusCode":404,"message":"Unable to update user","error":"Not Found"}
```

### Affected services

insights.opensauced.pizza

### Platforms

Desktop

### Browsers

Other (add additional context)

### Environment

Production

### Additional context

I'm using the Chromium based browser Arc, although.

### Code of Conduct

### Contributing Docs

doaortu:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

This PR fixes a bug caused by Twitter username input where entering the full Twitter URL caused an issue. Added a regex pattern to limit input to valid usernames with a maximum of 15 characters.

Fixes #1033 

https://user-images.githubusercontent.com/113872927/228781123-9486e1fd-0ff8-46a4-af94-1635f690b337.mp4

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/components/organisms/UserSettingsPage/user-settings-page.tsx
+++ b/components/organisms/UserSettingsPage/user-settings-page.tsx
@@ -216,6 +216,7 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
               classNames="bg-light-slate-4 text-light-slate-11"
               placeholder="saucedopen"
               label="Twitter Username"
+              pattern="^\w{1,15}$"
               name="twitter_username"
             />
             <TextInput


doaortu:
So, as far as I know, this fix also need to change the api.opensauced.pizza to run validation in the backend, should I fix that too?

doaortu:
So, I've added the ban words, but I met some complications:
The ban words must be case-insensitive because I've tried editing my username in twitter and it also case-insensitive there.
But, here's the catch:
- HTML input pattern does not support case-insensitive modifier, e.g: ```/word/i```
- Because the input pattern using javascript version of regex, it also doesn't support ```(?i)``` modifier, which supposed to enable me to add case-insensitivity in the negative lookahead clause.

This info based on [this stackoverflow answer](https://stackoverflow.com/a/5525111)

So, the only option is hardcode the case-insensitive negative lookahead for these words.


Summary, initially I go with:

```/^(?!.*?admin)(?!.*?twitter)\w{1,15}$/i```

but the i modifier can't be used in html input pattern,
then I end up hard code case-insesitivity like so:

```^(?!.*?[Aa][Dd][Mm][Ii][Nn])(?!.*?[Tt][Ww][Ii][Tt][Tt][Ee][Rr])\w{1,15}$```

What's your thoughts @0-vortex ?

0-vortex:
> So, I've added the ban words, but I met some complications:
> 
> The ban words must be case-insensitive because I've tried editing my username in twitter and it also case-insensitive there.
> 
> But, here's the catch:
> 
> - HTML input pattern does not support case-insensitive modifier, e.g: ```/word/i```
> 
> - Because the input pattern using javascript version of regex, it also doesn't support ```(?i)``` modifier, which supposed to enable me to add case-insensitivity in the negative lookahead clause.
> 
> 
> 
> This info based on [this stackoverflow answer](https://stackoverflow.com/a/5525111)
> 
> 
> 
> So, the only option is hardcode the case-insensitive negative lookahead for these words.
> 
> 
> 
> 
> 
> Summary, initially I go with:
> 
> 
> 
> ```/^(?!.*?admin)(?!.*?twitter)\w{1,15}$/i```
> 
> 
> 
> but the ```i``` modifier can't be used in html input pattern,
> 
> then I end up hard code case-insesitivity like so:
> 
> 
> 
> ```^(?!.*?[Aa][Dd][Mm][Ii][Nn])(?!.*?[Tt][Ww][Ii][Tt][Tt][Ee][Rr])\w{1,15}$```
> 
> 
> 
> What's your thoughts @0-vortex ?
> 
> 
> 
> Note: I also add the test cases for this regex in here: [https://regex101.com/r/e57vIt/6](https://regex101.com/r/e57vIt/6)

lowercase the string before checking the regex, solves a lot of problems 🫡🍕🔥

doaortu:
Okay I'll try that

OgDev-01 (approved):
LGTM 🍕

doaortu:
how about now? @0-vortex @OgDev-01 

I've extracted the pattern validation function to dedicated util function, I add custom validation messages too.

I've add some test too.

here's what it looks like now:

https://user-images.githubusercontent.com/113872927/229261019-48b8fc39-4d8b-4c96-8423-e0d9d937adf1.mp4

0-vortex (changes requested):
This is very good, couple of minor changes to get this shipped to production:
- refactor the validation type to an interface (make sure Validation keyword appears at the end)
- run npm run format to fix the missing semicolons

🔥❤️🍕

doaortu:
Thanks for the input, I'll fix it real quick

doaortu:
Changes: 
- changed ```TwitterValidationResult``` to ```TwitterUsernameValidation```
- changed ```type``` to ```interface```
- ran ```npm run format``` as suggested

Did I get it right?

0-vortex (approved):
LGTM! 👍 

@brandonroberts can you take a look? 👀

doaortu:
Thank you for accepting the PR.

I'm going to implement the same logic on the API.
But I have questions, how to run local development of the api.opensauced.pizza? I'm trying to do it myself but it always says missing supabaseUrl. 
I've tried build the docker using the Dockerfile, run docker compose up, renaming `.envrc` to `.env` but it always the same. I guess there's something missing steps in the `README.md` file.

0-vortex:
The backend databases, both dev and production are only accessible right now to @open-sauced/maintainers since they are hosted on digital ocean now; the env file required to connect is handed out securely but we could make the old supabase tables work if @brandonroberts and @bdougie agree.

In regards to the code change required, it would likely be shoving the regular expression you created in insights in a @Matches tag from class validator: https://github.com/typestack/class-validator

It would be replacing the (now) missing/incomplete twitter field in the user DTO to the same regex you created in the frontend with case insensitive flag or coupled with another @IsLowercase decorator.

Here's a screenshot from mobile: image

Another approach to shipping this without connecting to any data is through unit tests, however we are close to 0% coverage there and setting things up might be just as difficult. 

Since it's just a couple of lines to change, you might be able to do it anyway, while letting a reviewer test it out for you in a somewhat offline/delayed pairing session. 

Every contribution is appreciated! 🔥🙏🫡

bdougie:
My suggestion is not provide access to supabase or our dBs but create a way for folks to generate their own db locally.

I only have naive suggestions from.

1. Docker image that spins ups a Postgres with the need migrations and dummy data.
2. Use https://www.snaplet.dev

At this point it is up to someone outside of the opensauced core team to express the need and work with us to find the path.

Whatever the decision is, we should add the solution to the api readme and docs.opensauced.pizza

0-vortex:
> My suggestion is not provide access to supabase or our dBs but create a way for folks to generate their own db locally.
> 
> I only have naive suggestions from.
> 
> 1. Docker image that spins ups a Postgres with the need migrations and dummy data.
> 2. Use https://www.snaplet.dev
> 
> At this point it is up to someone outside of the opensauced core team to express the need and work with us to find the path.
> 
> Whatever the decision is, we should add the solution to the api readme and docs.opensauced.pizza

Investigated a couple of tools for this purpose with @brandonroberts in one of the engineering syncs, should definitely add snaplet to the list! We have to do a prod->dev seed every week, while also preparing multiple ETLs for elastic solutions. It's definitely e requirement on multiple ends! 🍕

doaortu:
Hi, thank you for the clarification and guidance.

I think, adding some kind of mock database would be nice for local development, so contributors outside opensauced core team can test and get immediate feedback.

as for this feature, I'm okay with this @0-vortex 's idea:

> Since it's just a couple of lines to change, you might be able to do it anyway, while letting a reviewer test it out for you in a somewhat offline/delayed pairing session.

I will submit the PR soon, but if it's not doable right now, feel free to discard it.

PR #1095, opened 2023-04-13
feat: User pages SSR & open graph

### LINKED ISSUE #879: Feature: Turn user profile pages into Server-side rendering to enable SEO & Microformats
### Type of feature

🍕 Feature

### Current behavior

Currently the user profile pages inner content are client-side rendered, and here's the illustration from google crawl bot:
image

The user data itself isn't server-side rendered, so can't be crawled and the SEO & microformats won't work on it.

### Suggested solution

Activate server-side rendering for these pages.

### Additional context

_No response_

### Code of Conduct

### Contributing Docs

- [ ] I agree to follow this project's Contribution Docs

### LINKED ISSUE #1091: Feature: Generate user profile image on demand when the url is shared
### Type of feature

🍕 Feature

### Current behavior

Screen Shot 2023-04-12 at 9 39 21 AM

### Suggested solution

Open graph images should hit `https://opengraph.opensauced.pizza/v1/users/bdougie`

https://graph.opensauced.xyz/users/bdougie.png

### Additional context

_No response_

### Code of Conduct

### Contributing Docs

- [ ] I agree to follow this project's Contribution Docs

nightknighto:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

- [x] Turns user pages `/user/[username]` into server-side rendering, instead of default static site generation.
- [x] Enables the meta tags in user pages.
- [x] Integrates opengraph service into meta tags.
- [x] Fetches the data in parallel. Time till finish on server on local decreased from ~800ms to ~400ms.

### Some Notes
- ~User profile's data is fetched on the server because it is needed for the user's bio meta tag. However, user's contribution data is not needed in the meta tags so it is fetched client-side to decrease the load on the server. This means a generic loading state for the whole page is now unacceptable, and there should be loading states over the components that wait for the contribution data. (Needs a ticket).~ Not applicable anymore till the below point is dealt with.
- The page's inner UI component was disabled from SSR as a quick fix for hydration issues. Results in a slower initial UI rendering for the profile. Needs a separate ticket to dive and investigate the cause of these issues.
- ~Requesting User data & social card data are done sequentially. They are independent and so should be done in parallel for faster response.~ Done.

Insight/Reasoning behind the decision to move to SSR: https://github.com/open-sauced/insights/issues/1091#issuecomment-1507501436

fixes #1091 
fixes #879 

image
image

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

Some tickets need opening, mentioned in the side-effects in the description.

### DIFF
--- a/pages/user/[username]/index.tsx
+++ b/pages/user/[username]/index.tsx
@@ -111,15 +111,10 @@ export async function handleUserSSR({ params }: GetServerSidePropsContext<{ user
   async function fetchSocialCardURL() {
     const socialCardUrl = `${String(process.env.NEXT_PUBLIC_OPENGRAPH_URL ?? "")}/users/${username}`;
     const ogReq = await fetch(`${socialCardUrl}/metadata`); //status returned: 204 or 304 or 404
-
-    if(ogReq.status === 404) { // image doesn't exist, wait for it to be generated
-      await fetch(socialCardUrl, {
-        method: "HEAD"
-      });
-    } else if(ogReq.status === 304) { // image exists but needs regenerating, no need to wait for the new version
+    if(ogReq.status !== 204) {
       fetch(socialCardUrl, {
         method: "HEAD"
-      });
+      }); // trigger the generation of the social card
     }
 
     ogImage = ogReq.headers.get("x-amz-meta-location");


pages/user/[username]/index.tsx
```diff
@@ -82,3 +83,35 @@ const Contributor: WithPageLayout = () => {
...
+  const req = await fetch(`${process.env.NEXT_PUBLIC_API_URL}/users/${username}`, {
+    headers: {
+      accept: "application/json",
+      Authorization: `Bearer ${sessionToken}`
+    }
+  });
+
+  const socialCardUrl = `${String(process.env.NEXT_PUBLIC_OPENGRAPH_URL ?? "")}/users/${username}`;
+  let ogImage = socialCardUrl;
+  const ogReq = await fetch(`${socialCardUrl}/metadata`); //statsu returned: 204 or 304 or 404
```
Comment (0-vortex) [inline pages/user/[username]/index.tsx:107]: this is not correct, `x-amz-meta-location` will always return the correct url, this endpoint is so the user knows they have to load the backend URL to generate the image. What you are doing here is changing the OG image url instead of working around the server side mechanism

0-vortex (changes requested):
It's not the correct image URL. Make a 2nd HEAD request for the actual image! 😅

pages/user/[username]/index.tsx
Comment (Deadreyo) [reply pages/user/[username]/index.tsx]: @0-vortex Hope this new commit correctly addresses the issue

pages/user/[username]/index.tsx
```diff
@@ -103,10 +103,10 @@ export async function handleUserSSR({ params }: GetServerSidePropsContext<{ user
   });
 
   const socialCardUrl = `${String(process.env.NEXT_PUBLIC_OPENGRAPH_URL ?? "")}/users/${username}`;
-  let ogImage = socialCardUrl;
-  const ogReq = await fetch(`${socialCardUrl}/metadata`); //statsu returned: 204 or 304 or 404
-  if(ogReq.status === 204 || ogReq.status === 304) {
-    ogImage = ogReq.headers.get("x-amz-meta-location") ?? socialCardUrl;
+  const ogReq = await fetch(`${socialCardUrl}/metadata`); //status returned: 204 or 304 or 404
```
Comment (0-vortex) [inline pages/user/[username]/index.tsx:106]:
```
let ogImage = socialCardUrl;
axios.get(`${socialCardUrl}/metadata`, {
      withCredentials: false,
      validateStatus: (status) => [204, 304, 404].includes(status)
    })
      .then(({ headers }) => {
        ogImage = headers.get("x-amz-meta-location");
        return axios.head(`${socialCardUrl}`, {
          withCredentials: false
        });
      });
```

and move the whole block after `const user = await req.json() as DbUser;` as the username would be validated in the database call

Comment (Deadreyo) [reply pages/user/[username]/index.tsx:106]: @0-vortex We don't have axios in the project. Do you want me to specifically install it or just do the equivalent with `fetch`?

pages/user/[username]/index.tsx
```diff
@@ -1,74 +1,75 @@
...
-
-  useEffect(() => {
-    Contributor.updateSEO!({
-      title: `${contributorLogin} | OpenSauced`,
-      description: `${user?.bio || `${profile?.githubName} has connected their GitHub but has not added a bio.`}`,
-      image: profile.githubAvatar,
-      twitterCard: "summary_large_image"
-    });
-  }, [contributorLogin, user?.bio, profile.githubAvatar]);
+  const githubAvatar = `https://www.github.com/${username}.png?size=300`;
```
Comment (brandonroberts) [inline pages/user/[username]/index.tsx:33]: Use `getAvatarByUsername` utility function

https://github.com/open-sauced/insights/blob/beta/lib/utils/github.ts#L6

pages/user/[username]/index.tsx
```diff
@@ -1,74 +1,75 @@
...
           />
         )}
       </Head>
       <div className="w-full">
-        <ContributorProfilePage
+        <ContributorProfilePageNoSSR
           prMerged={contributor[0]?.recent_merged_prs}
           error={isError}
-          loading={userLoading}
+          // loading={isLoading}
```
Comment (brandonroberts) [inline pages/user/[username]/index.tsx:66]: Remove commented-out code

```suggestion
```

pages/user/[username]/index.tsx
Comment (brandonroberts) [reply pages/user/[username]/index.tsx:106]: Just use fetch

brandonroberts (changes requested):

pages/user/[username]/index.tsx
Comment (0-vortex) [reply pages/user/[username]/index.tsx:106]: idea is you don't get the content back and don't need to await it, just generate the image with `method: "HEAD"`, linked axios code because it exists in typesync package and I don't know fetch equivalents 😅

Deadreyo:
@0-vortex @brandonroberts
Updated. The reason I was initially awaiting the generation request was to make sure the image finished generating before the the page is sent to the client, to avoid the case in which the page is sent and starts loading to the client but the image hasn't finished generating. Isn't this case possible?

Deadreyo:
Updated the code to now run the data fetching in parallel.

0-vortex:
> to avoid the case in which the page is sent and meta tags are loaded to the client but the image hasn't finished generating (so returning 404 in the social card). Isn't this case possible?

This is a "backfill" type (first time ever) generation issues, due to our S3 bucket being versioned: https://docs.aws.amazon.com/AmazonS3/latest/userguide/versioning-workflows.html

You are, however, correct that the fetch request can reject the promise under a network error case or the API returning 500 server error. In that case, we should catch and not let the meta generation fail or break the page, potentially returning a placeholder image.

0-vortex:
Some more information: image generation takes at best 7 seconds. When posting a user page to any social media, the 2nd server request will read the static re-generated.

You can test the caching needs update 304 status for users in the list with 3 days last generation, while, users not in this list will trigger 404:

Screenshot 2023-04-14 at 23 55 44

Deadreyo:
@0-vortex 
Alright how about this last commit as a solution for the image status? If the image doesn't exist (404), wait for it to be generated. If it is just outdated (304), then regenerate but don't wait (can use the old version).

0-vortex:
> @0-vortex
> Alright how about this last commit as a solution for the image status? If the image doesn't exist (404), wait for it to be generated. If it is just outdated (304), then regenerate but don't wait (can use the old version).

previous version was ok, this is not fixing the potential error.

Deadreyo:
@0-vortex Reverted.

bdougie (approved):
Let see if we can get this merged to move on to the next social card, Highlights. The approach and performance can be improved as we learn through more use cases.

brandonroberts (approved):

PR #1208, opened 2023-05-21
fix: redirect user to dashboard after logout on private pages

### LINKED ISSUE #1207: Bug: User is not redirected somewhere else on logout
@takanome-dev you shouldn't be on this page if you're not logged in. We have a bug that doesn't redirect you somewhere else on logout currently

_Originally posted by @brandonroberts in https://github.com/open-sauced/insights/issues/1180#issuecomment-1554594531_

takanome-dev:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description
This PR:
- adds a new wrapper component to check if the page is private or not
  - if the page is not private, do nothing
  - otherwise, check if the user is logged in. If it is, do nothing, and if it's not redirect to the dashboard
- adds a NextJS middleware file to check if the user is authenticated. The reason why this file is needed is that with the wrapper component, you can see the protected page for a second(s) before being redirected to the dashboard

Private pages where a user shouldn't have access without logging in:
- [x] /hub/*
- [x] /user/settings
- [x] /reports
- [x] private insight pages (pages/{username}/{insightId}/* when the `public` flag is set to `false` for the insight
- what else?

Fixes #1207 

Before

https://github.com/open-sauced/insights/assets/79809121/23b84cd1-4cee-47fe-8acd-54938dc74542

After 

https://github.com/open-sauced/insights/assets/79809121/05828607-541d-489e-838c-350f85d42a49

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook


brandonroberts:
Could this handle parts inside the app that require auth? 

Examples

- If you logout on the reports page, you still remain on the page, but you should be redirected somewhere
- Same for a private insight page

.github/workflows/triage.yml
```diff
@@ -15,6 +15,5 @@ jobs:
       - name: disable take an issue
         uses: bdougie/take-action@main
         with:
-          message: The auto-take issues workflow is not enabled here, we can only assign issues to @open-sauced/triage, @open-sauced/committers, @open-sauced/maintainers in this repo for now. We suggest folks not on the team open issues and take a look at the [Bugs](https://github.com/orgs/open-sauced/projects/6) or [Papercuts](https://github.com/open-sauced/insights/issues?q=is%3Aopen+is%3Aissue+milestone%3APapercuts). If you would like to find good first issues, please check out the companion project to this, [open-sauced/hot](https://github.com/open-sauced/hot).
```
brandonroberts, line 18: This should not be removed

takanome-dev:
> Could this handle parts inside the app that require auth?
> 
> Examples
> 
>     * If you logout on the reports page, you still remain on the page, but you should be redirected somewhere
> 
>     * Same for a private insight page

Yeah, if we set `isPrivateRoute` to true in the page, you will be redirected to the dashboard once logged out.
Currently, I have added that only on two pages: `/hub/*` and `/user/settings`

What other pages are private besides those? (feel free to edit this list above)(edit in progress)

.github/workflows/triage.yml
takanome-dev: There was a discussion with @bdougie about deleting this message [here](https://github.com/open-sauced/insights/issues/1192#issuecomment-1552930579)

brandonroberts:
@takanome-dev I updated the description with the use case for private insight pages

takanome-dev:
> I updated the description with the use case for private insight pages

@brandonroberts I was wrong about [this](https://github.com/open-sauced/insights/pull/1208#issuecomment-1557636378) answer :sweat_smile:, since it's inside the app, we need to use `useUser` hook from `@supabase/auth-helpers-react` to watch changes about the `user` data and redirect if the user logs out.

e2e-tests/UnsignedUsersRedirects.spec.ts
```diff
@@ -7,6 +7,5 @@ test("Homepage redirects to /javascript for unsigned users", async ({ page }) =>
 
 test("Insights Hub redirects to /javascript for unsigned users", async ({ page }) => {
   await page.goto("/hub/insights");
-  await expect(page).toHaveURL("/");
-  await expect(page).toHaveURL(/.*\/javascript.*/, { timeout: 10000 });
+  await expect(page).toHaveURL("http://localhost:3000/javascript/dashboard/filter/recent?redirectedFrom=%2Fhub%2Finsights", { timeout: 10000 });
```
brandonroberts, line 10: I don't think this should include `localhost`. How about just `/javascript/dashboard/filter/recent?redirectedFrom=%2Fhub%2Finsights`

brandonroberts (changes requested):

e2e-tests/UnsignedUsersRedirects.spec.ts
takanome-dev: That's right :+1:

bdougie:
There are unintended consequences to this change. You can not view the reports at all when not logged in. 

ezgif com-video-to-gif (4)

takanome-dev:
> There are unintended consequences to this change. You can not view the reports at all when not logged in.
> 
> ezgif com-video-to-gif (4)

Nice catch :+1: Will fix it

takanome-dev:
@bdougie Currently there is a bug in the reports page. If the user logs out and visits the reports page, he will see the upgrade button instead of signing in. 
We are checking for `hasReports === undefined` to force login but sometimes(from what I saw so far) `hasReports` is false, therefore the user can see the upgrade button.

reports-bug-1

If this bug is fixed, there won't be a need to redirect after logout (IMO they will be prompted to log in immediately)

How about fixing that in a separate pr?

brandonroberts:
@takanome-dev yes, let's fix in a follow-up so we can go ahead and land this

takanome-dev:
> @takanome-dev yes, let's fix in a follow-up so we can go ahead and land this

Sure, opened a pr for it https://github.com/open-sauced/insights/pull/1236

brandonroberts (approved):

PR #1218, opened 2023-05-24
feat: hide chatbot on non-prod environment and mobile screens

### LINKED ISSUE #1217: feat: Hide chat bot on non-production and desktop environments
The chat it should only show on "insights.opensauced.pizza" and not on localhost or deploy previews.

We should also remove it from mobile screens.

Code is in the the _app.tsx 

https://github.com/open-sauced/insights/blob/beta/pages/_app.tsx#L139-L141

image

takanome-dev:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

This PR hides the chatbot on non-production environments and on mobile screens.

> **Note**

```tsx
{!isMobile || NODE_ENV === 'production' || hostname === 'insights.opensauced.pizza' && (
  <Script id="siteGPT" type="text/javascript">
    ..
  </Script>
```

The code above won't work because:
- if you visit the app on desktop (first load), the chatbot will be there but won't go away when you switch on mobile (using devtools)
- if you visit the app on mobile (first load), the chatbot will be hidden and when you expand to desktop, it will show up. But if you go back to mobile, it will still be there.

The reason why that's happening is that, when the `script` is rendered, it will inject a button in the DOM, so even if you change the view to mobile and the script is hidden (like the code above), that won't help because the button is still in the DOM.

> **Note**

I have added an interval in the `useEffect` to check if the button is in the DOM because sometimes the script injects the chat button in the DOM after the first render.

Fixes #1217

hide-chatbot

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -56,6 +56,9 @@ function MyApp({ Component, pageProps }: ComponentWithPageLayout) {
   if (typeof window !== "undefined")
     hostname = window.location.hostname;
 
+  // JUST CHECKING WHAT THE HOSNAME WILL BE ON DEPLOY PREVIEW 😅
+  console.log({hostname});
+
   useEffect(() => {
     const chatWidget = document.getElementById("sitegpt-chat-icon");
     if (chatWidget) {
@@ -65,7 +68,7 @@ function MyApp({ Component, pageProps }: ComponentWithPageLayout) {
         chatWidget.style.display = "block";
       }
     }
-  },[isMobile, hostname, process.env.NODE_ENV]);
+  },[isMobile, hostname]);
 
   useEffect(() => {
     updateSEO(Component.SEO || {});
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -52,16 +52,20 @@ function MyApp({ Component, pageProps }: ComponentWithPageLayout) {
 
   const isMobile = useMediaQuery("(max-width:720px)");
 
+  let hostname = "";
+  if (typeof window !== "undefined")
+    hostname = window.location.hostname;
+
   useEffect(() => {
     const chatWidget = document.getElementById("sitegpt-chat-icon");
     if (chatWidget) {
-      if (isMobile || process.env.NODE_ENV !== "production") {
+      if (isMobile || process.env.NODE_ENV !== "production" || hostname !== "insights.opensauced.pizza") {
         chatWidget.style.display = "none";
       } else {
         chatWidget.style.display = "block";
       }
     }
-  },[isMobile]);
+  },[isMobile, hostname, process.env.NODE_ENV]);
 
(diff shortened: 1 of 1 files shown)


bdougie (changes requested):
We need to also check for the hostname is insights.opensauced.pizza. This feature is not needed on localhost or deploy previews.

```
const App = () => {
  const { hostname } = window.location;

  return (
    <>
      {hostname === 'insights.opensauced.pizza' && (
        <Script id="siteGPT" type="text/javascript">
          {"d=document;s=d.createElement(\"script\");s.src=\"https://sitegpt.ai/widget/365440930125185604.js\";s.async=1;d.getElementsByTagName(\"head\")[0].appendChild(s);"}
        </Script>
      )}
      {/* rest of the app */}
    </>
  );
};
```

OgDev-01:
A quick question @takanome-dev:  Why are we hiding the bot for mobile users?

takanome-dev:
> A quick question @takanome-dev: Why are we hiding the bot for mobile users?

That's what the issue states
> We should also remove it from mobile screens (https://github.com/open-sauced/insights/issues/1217#issue-1722801025)

OgDev-01:
> > A quick question @takanome-dev: Why are we hiding the bot for mobile users?
> 
> That's what the issue states
> 
> > We should also remove it from mobile screens ([#1217 (comment)](https://github.com/open-sauced/insights/issues/1217#issue-1722801025))

We can hide the tooltip and show just the bot button on mobile by setting `.tooltip-wrapper` class to `display: none ;` on mobile instead of removing the bot completely.

As a mobile user, I also have need for the chatbot 🤌

takanome-dev:
> As a mobile user, I also have need for the chatbot

Good point :+1: , thoughts @bdougie @brandonroberts?

bdougie:
We can keep mobile. I made the bubble smaller and removed the tooltip.

takanome-dev:
> It feels like this would be running every time you resize the screen by using the media query hook, which is not optimal. Can we ignore the check altogether after it has been hidden?

@brandonroberts Since we want to show the chatbot on mobile, the media query hook is not needed anymore :slightly_smiling_face: :

brandonroberts (approved):
LGTM

OgDev-01 (approved):

PR #1401, opened 2023-07-21
fix: Move card image api route

foxyblocks:
## Description

 I think we're running into this netlify caching issue for the static image endpoint. https://answers.netlify.com/t/netlify-function-with-query-strings-ignores-custom-cache-control-header/15390/1
I can reproduce it if i go to one of these URLs and then the other one.

- https://insights.opensauced.pizza/api/card.png?username=foxyblocks
- https://insights.opensauced.pizza/api/card.png?username=bdougie

This change moves the endpoint so that the username parameter is part of the path so the cache is correctly busted for different users.

I've yet to figure out how to access these parameters from the request other than doing string parsing. `request.nextUrl.searchParams.get('username')` works locally but doesn't work when deploying to netlify, probably due to some difference in runtime environment.

- [x] 🐛 Bug Fix

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/lib/utils/urls.ts
+++ b/lib/utils/urls.ts
@@ -19,7 +19,7 @@ export const siteUrl = (path: string = "") => {
  */
 export const cardPageUrl = (username: string) => siteUrl(`user/${username}/card`);
 
-export const cardImageUrl = (username: string) => siteUrl(`api/card.png?username=${username}`);
+export const cardImageUrl = (username: string) => siteUrl(`api/user/${username}/card.png`);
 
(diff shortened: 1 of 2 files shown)


brandonroberts (approved):

foxyblocks:
Should we merge this today or wait til monday?

PR #1459, opened 2023-08-01
feat: add github link to profile

a0m0rajab:
## Description

This adds a GitHub link to the OpenSauced User profile, it adds the link in two place, did that to have a conversation about it based on the previous PR: https://github.com/open-sauced/insights/pull/1449

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Fixes https://github.com/open-sauced/insights/issues/1199

This link in the profile page is redirecting to GitHub: 
image

To test:
- go to a user profile: https://insights.opensauced.pizza/user/a0m0rajab/highlights
- click on the mention (tag) under their profile
- Check if it's redirecting to the GitHub

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/components/molecules/ContributorProfileInfo/contributor-profile-info.tsx
+++ b/components/molecules/ContributorProfileInfo/contributor-profile-info.tsx
@@ -96,7 +96,7 @@ const ContributorProfileInfo = ({
               <span className="flex gap-2 items-center">
                   <FiGithub className="text-light-slate-9" />
                   <Link
-                    href={`https://twitter.com/${twitterUsername}`}
+                    href={`https://github.com/${githubName}`}
                     target="_blank"
                     rel="noreferrer"
                     className="w-max hover:text-orange-500 "
@@ -109,7 +109,7 @@ const ContributorProfileInfo = ({
                 <span className="flex gap-2 items-center">
                   <FiTwitter className="text-light-slate-9" />
                   <Link
-                    href={`https://github.com/${githubName}`}
+                    href={`https://twitter.com/${twitterUsername}`}
                     target="_blank"
                     rel="noreferrer"
                     className="w-max hover:text-orange-500 "
--- a/components/molecules/ContributorProfileInfo/contributor-profile-info.tsx
+++ b/components/molecules/ContributorProfileInfo/contributor-profile-info.tsx
@@ -3,7 +3,7 @@ import Link from "next/link";
 
 import { AiOutlineGift } from "react-icons/ai";
 import { BsDiscord } from "react-icons/bs";
-import { FiClock, FiGithub, FiLinkedin, FiTwitter } from "react-icons/fi";
+import { FiClock, FiGithub, FiLinkedin, FiTwitter, FiGithub} from "react-icons/fi";
 
 import clsx from "clsx";
 import LanguagePill from "components/atoms/LanguagePill/LanguagePill";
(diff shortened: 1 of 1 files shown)


components/molecules/ContributorProfileInfo/contributor-profile-info.tsx
```diff
@@ -54,7 +54,7 @@ const ContributorProfileInfo = ({
           {isConnected && <Badge text="beta" />}
         </div>
         <div className="flex items-center text-sm gap-3">
-          <span className="text-light-slate-11 text-sm">{`@${githubName}`}</span>
+          <a className="text-light-slate-11 text-sm" href={`https://github.com/${githubName}`}>{`@${githubName}`}</a>
```
babblebey, line 57:
```suggestion
          <span className="text-light-slate-11 text-sm">{`@${githubName}`}</span>
```
I don't think the `@[username]` should link out to GitHub, because visually it doesn't look like something that should link to GitHub, I think its best for User Experience to either have this just link back to the same profile on open-sauced/insights or stay without a link at all.

The `GitHub` button link in the Socials Section below the About section, unlike the `@[username]` tells me exactly what/where I'm expecting to land when I click, and its feels just enough.

NsdHSO: Why don't we use the `Link` component instead of using `a` directly

OgDev-01: This was done on purpose so we can discuss the better one that fits our needs. The GitHub link below the about section pretty much fits our needs here. So we can move on with this 👍

OgDev-01: > Why don't we use the `Link` component instead of using `a` directly

The `Link` tag does not support external URLs in a new tab

OgDev-01: Discard my previous comment... Just found out that links can be used in this case since Next13 https://stackoverflow.com/a/74382873 :<

bdougie: I might have a counter point similar to @babblebey. I might argue that the `@[username]` is not needed. The GitHub profile being linkable below is fine and we are not using real names on the platform so the username is already repeated multiple places.

a0m0rajab: would this be a better design: 
image

a0m0rajab: @babblebey, @NsdHSO thank you for your feedback!

a0m0rajab:
The after designs: 
image

https://deploy-preview-1459--oss-insights.netlify.app/user/bdougie/highlights

components/molecules/ContributorProfileInfo/contributor-profile-info.tsx
bdougie: > would this be a better design: 
> 
> image
> 
> 

This is much better me

a0m0rajab: This is the current implementation, 
image

Side note, I noticed the image you added after writing my comment, the page did not load the edit.

bdougie (approved):

brandonroberts (approved):

PR #1492, opened 2023-08-05
fix: add space between avatar and  cta menu

### LINKED ISSUE #1491: Bug: no space between the buttons and the profile picture
### Describe the bug

In the insights user profile on mobile phone, the share and get card button are squashed with the user profile picture
Here's a screenshot
Screenshot_20230805-122449

### Steps to reproduce

1. Go to https://insights.opensauced.pizza/user/dev-phantom on mobile 
2. See error 

### Browsers

Chrome

_No response_

### Code of Conduct

### Contributing Docs

DevPres:
## Description

This PR adds icon buttons to the ContributorProfileHeader component for mobile breakpoints, addressing spacing issues.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [x] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Fixes #1491 

Before 

image

After 

image

## Added tests?

- [ ] 👍 yes
- [ x] 🙅 no, because they aren't needed

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook
- [ x] 🙅 no documentation needed

### DIFF
--- a/components/molecules/ContributorProfileHeader/contributor-profile-header.tsx
+++ b/components/molecules/ContributorProfileHeader/contributor-profile-header.tsx
@@ -146,17 +146,29 @@ const ContributorProfileHeader = ({
             {/* Mobile dropdown menu */}
 
             <DropdownMenu>
-              <div className="flex items-center gap-2 mb-10 md:gap-6 flex-wrap">
-                <Button variant="primary" href={cardPageUrl(username!)}>
+              <div className="flex items-center md: gap-2 mb-10 md:gap-6 flex-wrap">
+                <Button className="sm:hidden" variant="primary" href={cardPageUrl(username!)}>
+                  <FaIdCard className="" />
+                </Button>
+                <Button className="hidden sm:inline-flex" variant="primary" href={cardPageUrl(username!)}>
                   <FaIdCard className="mt-1 mr-1" /> Get Card
+                </Button>
+
+                <Button
+                  onClick={() => handleCopyToClipboard(`${host}/user/${user?.user_metadata.user_name}`)}
+                  className="bg-white sm:hidden"
+                  variant="text"
+                >
+                  <FiCopy className="" />
                 </Button>
                 <Button
                   onClick={() => handleCopyToClipboard(`${host}/user/${user?.user_metadata.user_name}`)}
-                  className="px-8 py-2 bg-white "
+                  className="px-8 py-2 bg-white hidden sm:inline-flex"
                   variant="text"
                 >
                   <FiCopy className="mt-1 mr-1" /> Share
                 </Button>
+
                 {!isOwner && (
                   <DropdownMenuTrigger title="More options" className="p-2 mr-3 bg-white rounded-full cursor-pointer ">
                     <TfiMoreAlt size={20} className="" />
--- a/components/molecules/ContributorProfileHeader/contributor-profile-header.tsx
+++ b/components/molecules/ContributorProfileHeader/contributor-profile-header.tsx
@@ -118,7 +118,7 @@ const ContributorProfileHeader = ({
         </div>
       )}
 
-      <div className="container flex flex-row items-end justify-between px-2 py-6 mx-auto md:px-16">
+      <div className="container flex flex-row items-end justify-between gap-2 px-2 py-6 mx-auto md:px-16">
         <div className="translate-y-[65px] hidden md:inline-flex">
           <Avatar
             initialsClassName="text-[100px] -translate-y-2.5  leading-none"


bdougie:
Thanks for the contribution. I don't think this is enough of a fix. Rather than simply add space to the buttons, I think use should get a different interaction. 

Previously we tested a dropdown.

DevPres:
i can continue working on this branch or do i have to open new one?

bdougie:
@DevPres could we move the `share` link to the dropdown for now and solve tis problem all together?

Screen Shot 2023-08-08 at 8 21 00 AM

DevPres:
Hi @bdougie ,

I did some changes in my PR, and changed the description with new screenshots.
Let me know if it's ok in this way

bdougie:
Adding the button and share in the drop-down is not needed.

Just the word share is fine. 

Also noticed the original issue of spacing is not resolved either. Each example is quite different and spacing.

Please consider the user experience with this. What is the goal for the user on this page while on mobile? 

My answer is to quickly share and view a profile. 

An alternative to the drop-down could be just displaying icons without text. We had this as an option months ago when viewing github profiles were the only options.

bdougie:
@DevPres please proceed with just icons. The icons have enough context for the user. 

> However, I didn't quite understand why you're saying the spacing issue wasn't resolved. Indeed, now the buttons no longer overlap the avatar, which was the initially reported bug.

mostly mention due to scope creep from this, but it was necessary to get to a solution.

DevPres:
Perfect!

I' ll work on this during this days

DevPres:
Hi @bdougie  !

I pushed a version with only icons button. i updated the PR description.

bdougie (approved):
This is much better

NsdHSO (approved):
LGMTM

components/molecules/ContributorProfileHeader/contributor-profile-header.tsx
```diff
@@ -146,17 +146,29 @@ const ContributorProfileHeader = ({
...
             <DropdownMenu>
-              <div className="flex items-center gap-2 mb-10 md:gap-6 flex-wrap">
-                <Button variant="primary" href={cardPageUrl(username!)}>
+              <div className="flex items-center md: gap-2 mb-10 md:gap-6 flex-wrap">
+                <Button className="sm:hidden" variant="primary" href={cardPageUrl(username!)}>
+                  <FaIdCard className="" />
+                </Button>
+                <Button className="hidden sm:inline-flex" variant="primary" href={cardPageUrl(username!)}>
                   <FaIdCard className="mt-1 mr-1" /> Get Card
                 </Button>
```
babblebey, line 155:
```suggestion
                <Button variant="primary" href={cardPageUrl(username!)}>
                  <FaIdCard className="mt-1 sm:mr-1" /> 
                  <span className="hidden sm:inline-flex">Get Card<span/>
                </Button>
```

You could have a responsive `Button` instead of rendering 2 separate buttons that does the same thing.

components/molecules/ContributorProfileHeader/contributor-profile-header.tsx
```diff
@@ -146,17 +146,29 @@ const ContributorProfileHeader = ({
...
+                  <FiCopy className="" />
+                </Button>
                 <Button
                   onClick={() => handleCopyToClipboard(`${host}/user/${user?.user_metadata.user_name}`)}
-                  className="px-8 py-2 bg-white "
+                  className="px-8 py-2 bg-white hidden sm:inline-flex"
                   variant="text"
                 >
                   <FiCopy className="mt-1 mr-1" /> Share
                 </Button>
```
babblebey, line 170: You can replicate the responsiveness for this `Button` too, instead of a second button.

babblebey (commented):
Looks good @DevPres 

But I have this quick suggestion.

components/molecules/ContributorProfileHeader/contributor-profile-header.tsx
DevPres: Hi @babblebey ,
I had tried your solution, but unfortunately with the `span`, I would have had to reapply the CSS classes on that as well. The problem with this approach is that if there were changes to the design in the future, such as the text of the buttons becoming bold, it would break here, and someone would have to modify this component as well.

To avoid button duplication, one could consider using a JavaScript method to determine the current breakpoint. However, this would deviate from the focus of issue.

PR #1513, opened 2023-08-07
revert: "fix: Feed page responsiveness" URGENT!!!

babblebey:
Reverts open-sauced/insights#1490

We have a regression detected and some parts to review see https://github.com/open-sauced/insights/pull/1490#pullrequestreview-1565363428


brandonroberts:
@babblebey why do we need to revert instead of fixing forward?

bdougie:
> @babblebey why do we need to revert instead of fixing forward?

going to mention the same. Beta is not live to anyone except other contributors.

babblebey:
Oh well noted 😃

First time experiencing this, I shook and clicked revert 🤦🏾‍♂️

@RitaDee is currently on it!

PR #1581, opened 2023-08-20
feat(filter-header): implement search for custom topic in Explore page

Brian-Pob:
## Description

This PR implements custom topic search on the explore page.

Added search component to filter-header for searching for a custom topic that is not in the topics list. Can definitely benefit from additional tests related to the search bar. Styling and functionality, especially on initial page load, may still be improved.

Suggestions and criticisms are welcome!

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Implements open-sauced/feedback#24 

old ui

showing new ui with search bar

## Added tests?

- [ ] 👍 yes

Not familiar enough with testing to know how to implement tests related to the custom topic search.

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/components/molecules/FilterHeader/filter-header.tsx
+++ b/components/molecules/FilterHeader/filter-header.tsx
@@ -1,5 +1,6 @@
 import { useRouter } from "next/router";
 
+import { useEffect, useState } from "react";
 import Text from "components/atoms/Typography/text";
 import Title from "components/atoms/Typography/title";
 import ContextThumbnail from "components/atoms/ContextThumbnail/context-thumbnail";
@@ -13,6 +14,7 @@ import FilterCardSelect from "components/molecules/FilterCardSelect/filter-card-
 import getTopicThumbnail from "lib/utils/getTopicThumbnail";
 import { interestsType } from "lib/utils/getInterestOptions";
 import { getInterestOptions } from "lib/utils/getInterestOptions";
+import Search from "components/atoms/Search/search";
 
 const HeaderFilter = () => {
   const router = useRouter();
@@ -20,7 +22,9 @@ const HeaderFilter = () => {
   const topicOptions = getInterestOptions();
 
   const { filterValues } = useFilterPrefetch();
-  const { filterName, toolName, selectedFilter } = router.query;
+  const { toolName, selectedFilter } = router.query;
+  const [filterName, setFilterName] = useState<string | string[] | undefined>(router.query.topicName);
+
   const filterBtnRouting = (filter: string) => {
     captureAnayltics("Filters", "toolsFilter", `${filter} applied`);
     return router.push(`/${filterName}/${toolName}/filter/${filter.toLocaleLowerCase()}`);
@@ -31,8 +35,15 @@ const HeaderFilter = () => {
   };
 
   const topicRouting = (topic: string) => {
+    if (!topic) return;
+    setFilterName(topic);
     router.push(`/${topic}/${toolName}${selectedFilter ? `/filter/${selectedFilter}` : ""}`);
   };
+
+  useEffect(() => {
+    // If filterName is null, get the value from the url
+    setFilterName(document.location.pathname.split("/")[1] as string);
+  }, []);
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


components/molecules/FilterHeader/filter-header.tsx
```diff
@@ -13,14 +14,18 @@ import FilterCardSelect from "components/molecules/FilterCardSelect/filter-card-
...
 const HeaderFilter = () => {
   const router = useRouter();
   const filterOptions = useFilterOptions();
   const topicOptions = getInterestOptions();
 
   const { filterValues } = useFilterPrefetch();
-  const { filterName, toolName, selectedFilter } = router.query;
+  const { toolName, selectedFilter } = router.query;
+  const [filterName, setFilterName] = useState<string | string[] | undefined>(router.query.filterName);
+  console.log("filterName 1", filterName);
```
NsdHSO, line 27: nit: Please could you remove this line, or is it an intentional feature ???
```suggestion
```

Brian-Pob: My bad. It was left over from debugging. I'll remove it in the morning.

NsdHSO (commented):
Tnx for your PR, could you make these changes ?? And this PR IMHO is Good to ship.

Brian-Pob:
> Tnx for your PR, could you make these changes ?? And this PR IMHO is Good to ship.

Just added the changes. And thank you!

bdougie:
Functionality this is ok, but this is only helpful for folks who have a good understanding of our product. The search box and the dropdown seems confusing. 

I think this would be better fit as part of the dropdown. I am sure there is a design system we can borrow from. i.e. [radix](https://www.radix-ui.com/), [shadcn](https://ui.shadcn.com/), or similar. 

Screen Shot 2023-08-21 at 12 43 09 PM

Brian-Pob:
I agree. I thought about including the search functionality as part of the drop down at first, but I saw that the drop down atom was used in other places. I wasn't sure if it was alright for me to edit an atomic component in such a way that it might affect its use in other parts of the project.

I have ideas for what to implement that are better in terms of UI/UX. I just need a bit of help in the actual implementation.

bdougie:
>I wasn't sure if it was alright for me to edit an atomic component in such a way that it might affect its use in other parts of the project.
> 

We don't have strong rules outside of atomic design. My thought would be create a new atom that is a dropdown-with-search

Brian-Pob:
It looks like integrating the search functionality into the dropdown is more complicated than I assumed.

The molecule associated with the dropdown is called `filter-card-select.tsx` and it is actually a [`<Select>`](https://www.radix-ui.com/themes/docs/components/select) element, not a [`<DropdownMenu>`](https://www.radix-ui.com/themes/docs/components/dropdown-menu). 

A Select is meant to allow the user to choose a `<SelectItem>` from a pre-configured list, which is what it currently does. A DropdownMenu is more flexible in terms of what can be included in it.

Adding the search bar to the Select adds several complications
1. The search bar cannot be navigated to using the keyboard (tab and arrow keys cannot select the search bar).
2. Typing into the search bar will trigger the built-in keyboard listener of the Select. For example, if you want to type "css," typing "c" will remove focus from the search bar and then highlight the first SelectItem that begins with "c." In this case, that is "csharp."

GIF demonstrating problems with attempt at implementing search inside the Select element


3. Wrapping the search bar inside a `<SelectItem>` tag introduces more problems that I won't get into for the sake of brevity. 

I have made a separate [branch](https://github.com/Brian-Pob/insights/tree/select-search) with my attempts to implement Search this way.

---

I have thought of two potential paths to take. Of course, these are not the only options. These are just what I have thought of.
1. Improve the UX of the PR as it is currently. We could play around with the behavior of the search bar, labels, positioning, etc. to see how we could make its functionality as clear as possible.
2. Refactor `filter-card-select.tsx` to use a `<DropdownMenu>` instead of a `<Select>`. This would be the most hands-on approach and would involve the most code changes.

Personally, I think Option 1 would be the simplest and most effective option.

I am still very much open to any suggestions and guidance.

Brian-Pob:
Closing this pull request as this feature should have further discussion on how best to implement it. May reopen in the future but would probably be better off making a new PR.

PR #1613, opened 2023-08-25
feat: profile tabs - path params to query params (1229)

### LINKED ISSUE #1229: Feature: refactor user profile tabs routing implementation
### Type of feature

🍕 Feature

### Current behavior

We are currently appending the tabs to the url path 

### Exp:
- highlights -> `/user/{username}/highlights`
- contributions -> `/user/{username}/contributions`

This solution works... But if we are to put SEO into consideration, A better solution is using [URL parameters](https://www.botify.com/learn/basics/what-are-url-parameters)

### Exp:

- highlights -> `/user/{username}?tab=highlights`
- contributions -> `/user/{username}?tab=contributions`

### What are the benefits of this approach?

- **SEO and shareability**: URL parameters are more search engine friendly and shareable. Search engines can index individual tab URLs, and users can easily copy and share specific tab URLs.
- **Browser history**: URL parameters modify the URL itself, creating a new entry in the browser history. This allows users to bookmark or navigate back to specific tab URLs.
- **Server-side rendering**:  URL parameters are easier to handle as they can be processed on the server and the appropriate tab can be rendered accordingly.

### Suggested solution

Change the current path implementation to urlParams to handle all tabs in the user profile page

### Additional context

_No response_

### Code of Conduct

### Contributing Docs

ThePiyushAggarwal:
## Description

This PR replaces the use of path parameters in user profile tabs and instead uses query parameters. Closes #1229 

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Closes #1229 

| Before | After |
| ---- | ---- |
| image | image |

Video

[profile-tabs-query-params-usage.webm](https://github.com/open-sauced/app/assets/27003616/c1b7fcce-9d39-45e8-b2e1-01dfb1d60400)

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook


ThePiyushAggarwal:
Hi @OgDev-01 !

Can you please help me with this PR? I have two solutions for one problem. Both of them work well but I need confirmation from maintainers for go ahead.

The problem - Many users would have shared the URL to on their profiles, we need to properly redirect them to the new URL with query parameters.

For commit 2fc4d4a, I have made use of redirection support in `next.config.ts`. This seemed like a better solution since the `user/[username]/highlights` paths don't exist anymore. And I can safely delete the files like `highlights.tsx`, `contributions.tsx`... But deleting those files also seem wrong in some way.

I worked on another solution also. That involves me doing something like this in `highlights.tsx`, `contributions.tsx`... and other files.
```ts
import { TabKey } from "components/organisms/ContributorProfileTab/contributor-profile-tab";
import { handleUserSSR, UserSSRPropsContext } from "../[username]";

export default function Highlights() {
  return <></>;
}

export const getServerSideProps = async (context: UserSSRPropsContext) => {
  if (context.params?.username) {
    return {
      redirect: {
        destination: `/user/${context.params.username}?tab=${"highlights" satisfies TabKey}`,
        permanent: true,
      },
    };
  }

  return await handleUserSSR(context);
};
```

I did this edit in all four files and it was working the same.

Please let me know your thoughts. Thank you for your support!

components/organisms/ContributorProfileTab/contributor-profile-tab.tsx
```diff
@@ -57,78 +69,47 @@ const ContributorProfileTab = ({
...
-    if (login && currentPathname) {
-      setCurrentPathname(getCurrentPathName);
-      router.push(`/user/${login}/${currentPathname}`);
-    }
-    if (login && !hasHighlights && currentPathname) {
-      setCurrentPathname(currentPathname);
-      router.push(`/user/${login}/${currentPathname}`);
-    }
-  }, [login, hasHighlights]);
+  const hasHighlights = highlights ? highlights.length > 0 : false;
```
OgDev-01, line 75: `hasHighlights` needs to stay as a state because the highlight data that updates it is an asynchronous call that might take some time to complete, hence the need to update the `hasHighlights` state from a useEffect.

OgDev-01 (changes requested):

components/organisms/ContributorProfileTab/contributor-profile-tab.tsx
OgDev-01: As far as implementation goes, This looks good 🍕. We just have to check on the initial page mount to update the tab query in the URL painted 👍.

ThePiyushAggarwal: Thank you for your suggestion. However, as per my understanding, using a state variable for `hasHighlights` may not be necessary in this case because the value of `hasHighlights` is derived from the value of `highlights`, which is returned by the `useSWR` hook. When the value of `highlights` changes, the value of `hasHighlights` will also be updated automatically.

Here's a video of this happening in this feature with logs.

[highlights-hasHighlights-state-issue.webm](https://github.com/open-sauced/app/assets/27003616/0d38e46e-c0b2-43e9-9339-b3b065d14725)

Here's a link for the a 5 min [YouTube chapter](https://youtu.be/GGo3MVBFr1A?t=505&si=P8H_CWjT3G2pKrse). Hope you like it and understand what I am trying to do here.

OgDev-01: Thanks for sharing this light 🤗🤗🤗... Now I understand👍.

I think my second comment is a case we just have to handle then 👌

>  We just have to check on the initial page mount to update the tab query in the URL painted 👍.

ThePiyushAggarwal: I am glad you watched it. It was a good insight. I am taking care of the other suggestion.

ThePiyushAggarwal: I have made changes according to the suggestion and marked the PR as Ready for Review.

OgDev-01 (approved):
Works as espected! 🍕

jpmcb (approved):
Seems to work as expected: I'll leave it up to others to comment on the React 👍🏼

bdougie:
Can you fix this in your PR @OgDev-01. In addition to 404 for not found, the link to profiles all default to highlights. 

Rewatch GIF Recording - 2023-08-29 at 7 33 35 AM

OgDev-01:
> the link to profiles all default to highlights.

Yes, this will be handled [here](https://github.com/open-sauced/app/pull/1631)

PR #1648, opened 2023-09-01
fix: add check for undefined tagged_repos

### LINKED ISSUE #1644: Bug: User profile page throws Application error (cannot read property of undefined: reading map)
### Describe the bug

image

### Steps to reproduce

1. Go to  https://insights.opensauced.pizza/user/diivi

### Browsers

Chrome

_No response_

### Code of Conduct

### Contributing Docs

brandonroberts:
## Description

- Adds a fallback check for `tagged_repos` in the event that the cache has highlight data that hasn't been invalidated with the latest release.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Closes #1644

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/components/molecules/ContributorHighlight/contributor-highlight-card.tsx
+++ b/components/molecules/ContributorHighlight/contributor-highlight-card.tsx
@@ -96,7 +96,7 @@ const ContributorHighlightCard = ({
 }: ContributorHighlightCardProps) => {
   const { toast } = useToast();
 
-  const formattedTaggedRepos = taggedRepos.map((repo) => {
+  const formattedTaggedRepos = (taggedRepos || []).map((repo) => {
     const [repoOwner, repoName] = repo.split("/");
     const repoIcon = getAvatarByUsername(repoOwner, 60);
     return { repoName, repoOwner, repoIcon };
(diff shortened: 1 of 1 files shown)


jpmcb (approved):
👀 So that tagged repo data wasn't being sent from the API or was somehow invalidated?

bdougie (approved):

diivi (approved):

brandonroberts:
> 👀 So that tagged repo data wasn't being sent from the API or was somehow invalidated?

We use a client-side cache with [SWR](https://swr.vercel.app/) that has highlight data in it. It could be that the cache was still present without the `tagged_repos` property in the data, causing the array to be undefined. It explains why it works with an incognito window because there's no existing cache to pull from.

I don't know if this is a common pattern with SWR or not, but maybe we should look into creating a unique cache key with each release, so we can better guarantee old data isn't being used.

jpmcb:
> I don't know if this is a common pattern with SWR or not, but maybe we should look into creating a unique cache key with each release, so we can better guarantee old data isn't being used.

Ahhh I see. Bummer: when I've seen that pattern, the key that's used is a string representation of the API which is typically "hard" semvered. So querying `useSWR('/v123/api/user', fetcher)` always returns results for that specific version of the API that coresponds to the version of the frontend app. But since we're not really supporting multiple versions of the API that correspond to specific versions of the insights app, that doesn't really make sense. Maybe the key just needs to be the version of the app that the user has and that corresponds to the data that is fetched from the cache.

PR #1697, opened 2023-09-11
Move TooltipProvider to storybook global decorator

foxyblocks:
## Description

- Moves the @radix-ui TooltipProvider into a global storybook decorator. This will allow the creation of new stories without worrying about wrapping them in the provider and updates all stories that were previously using it to no longer reference it directly. I noticed that several of these stories didn't even need the provider but were still depending on it. 
- Converts the `storybook/preview.js` file to typescript so the the exported config object can be typechecked. 
- every story that was touched was also updated to the latest storybook CF3 component format which is much terser. 

- [x] 🧑‍💻 Code Refactor

none

none

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [x] 📕 storybook


OgDev-01 (approved):
Thanks for this fix @foxyblocks, that tooltip provider has been a pain for sometime now😇

brandonroberts:
Thanks @foxyblocks. We said we were going to revisit this at some point

PR #1707, opened 2023-09-18
feat: add feature flag integration

brandonroberts:
## Description

- Adds feature flag integration using tier.run
- Updates the `.env` with a limited access read-only token and endpoint to tier.run
- Connects the reports feature to the feature access checks

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/lib/utils/feature-access.ts
+++ b/lib/utils/feature-access.ts
@@ -1,0 +1,30 @@
+import { Limits } from "tier";
+
+function hasAccess(access: Limits["usage"], featureName: string) {
+  return !!access.find((limit) => limit.feature === `feature:${featureName}` && limit.used < limit.limit);
+}
+
+const getFeatureAccess = async () => {
+  const limits = await loadAccess();
+  const hasReports = hasAccess(limits, "reports");
+
+  return {
+    hasReports,
+  };
+};
+
+const loadAccess = async () => {
+  const response = await fetch(`/api/session`, {
+    method: "GET",
+  });
+
+  if (response.ok) {
+    return (await response.json()).access as Usage;
+  }
+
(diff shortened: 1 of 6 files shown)


lib/utils/feature-access.ts
```diff
@@ -0,0 +1,30 @@
...
+const loadAccess = async () => {
+  const response = await fetch(`/api/session`, {
+    method: "GET",
+  });
+
+  if (response.ok) {
+    return (await response.json()).access as Usage;
+  }
+
+  return [] as unknown as Usage;
```
jpmcb, line 25: I'm guessing if there's a problem getting the session data we'll throw an error somewhere else in the app? Or do we need to handle it here?

jpmcb (approved):

lib/utils/feature-access.ts
brandonroberts: It would be somewhere else. If there is no feature access data available, this would fallback to very limited access for our current checks (reports).

OgDev-01 (approved):
Totally new to me, but code looks Good 👍

bdougie (approved):
🚀

PR #1734, opened 2023-09-21
feat: added contributor page for lists

### LINKED ISSUE #1710: Feature: Add contributors page for a list
### Type of feature

🍕 Feature

### Current behavior

When a user opens a list and clicks on the contributors tab, we want to display a paginated list of the contributors

image

### Suggested solution

- From the `/hub/lists` page, select an item from the list
- Click on the `Contributors` tab. The URL is `/lists/{listId}/contributors`
- The contributors table is displayed for the contributors added to that particular list

### Additional context

_No response_

### Code of Conduct

### Contributing Docs

- [x] I agree to follow this project's Contribution Docs

nickytonline:
## Description

This add the contributors list page. If you want to try it:

1. Grab this PR, e.g. `gh co 1734`
2. Run `npm install`
3. Run `npm run dev`
4. Navigate to http://localhost:3000/lists/2afc5e31-2925-4030-8818-abfaecdbd174/contributors (the list is public)
5. You can also add your own list then navigate to it, http://localhost:3000/hub/lists/new then navigate to http://localhost:3000/lists/your_list_id/contributors

https://github.com/open-sauced/app/assets/833231/ac22edea-388e-46ee-a23c-d17889b7cf9a

* I mention this in the comments, but I wrapped the component with the `<ListPageLayout />` instead of setting it via `ContributorsListPage.Layout = ListPageLayout` because there was data it needed that I couldn't get without two calls on the client-side pre-SSR changes. Now it can receive props. This probably makes sense for other pages as well going forward, but it can be done on a per page basis.
* The first load of the page will result in a hydration error (working on it). Close the error modal then you can see things in action.
* As discussed in the engineering sync, the checkbox column is not available for this iteration, and the edit button (if you're the list owner) is not available either.

Left to do:
* Tests
* Still sorting out the hydration issue

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Closes #1710

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/components/ListHeader/list-header.tsx
+++ b/components/ListHeader/list-header.tsx
@@ -16,9 +16,10 @@ interface ListHeaderProps {
   listId: string;
   isPublic: boolean;
   isOwner: boolean;
+  numberOfContributors: number;
 }
 
-const ListHeader = ({ name, isPublic, listId, isOwner }: ListHeaderProps): JSX.Element => {
+const ListHeader = ({ name, isPublic, listId, isOwner, numberOfContributors }: ListHeaderProps): JSX.Element => {
   const { toast } = useToast();
   const posthog = usePostHog();
 
@@ -48,7 +49,7 @@ const ListHeader = ({ name, isPublic, listId, isOwner }: ListHeaderProps): JSX.E
             {<Badge isPublic={isPublic} />}
           </div>
           <div className="flex items-center gap-2 mt-4" id="contributorCount">
-            Contributors
+            {numberOfContributors} Contributors
           </div>
         </div>
       </div>
--- a/components/ListHeader/list-header.tsx
+++ b/components/ListHeader/list-header.tsx
@@ -44,7 +44,7 @@ const ListHeader = ({ name, isPublic, listId, isOwner, numberOfContributors }: L
         <div className="flex flex-col justify-center p-2 header-info grow">
           <div className="flex gap-2">
             <Title level={1} className="!text-2xl font-semibold tracking-tight text-slate-900">
-              {(name && truncateString(name, 30)) || "Insights"}
+              {(name && truncateString(name, 30)) || "List"}
             </Title>
             {<Badge isPublic={isPublic} />}
           </div>
--- a/components/ListHeader/list-header.tsx
+++ b/components/ListHeader/list-header.tsx
@@ -12,12 +12,13 @@ import { truncateString } from "lib/utils/truncate-string";
 import { useToast } from "lib/hooks/useToast";
 
 interface ListHeaderProps {
-  list?: DbUserList;
+  name: string;
   listId: string;
+  isPublic: boolean;
   isOwner: boolean;
 }
 
-const ListHeader = ({ list, listId, isOwner }: ListHeaderProps): JSX.Element => {
+const ListHeader = ({ name, isPublic, listId, isOwner }: ListHeaderProps): JSX.Element => {
   const { toast } = useToast();
   const posthog = usePostHog();
 
@@ -42,11 +43,13 @@ const ListHeader = ({ list, listId, isOwner }: ListHeaderProps): JSX.Element => 
         <div className="flex flex-col justify-center p-2 header-info grow">
           <div className="flex gap-2">
             <Title level={1} className="!text-2xl font-semibold tracking-tight text-slate-900">
-              {(list && truncateString(list.name, 30)) || "Insights"}
+              {(name && truncateString(name, 30)) || "Insights"}
             </Title>
-            {list && <Badge isPublic={list?.is_public} />}
+            {<Badge isPublic={isPublic} />}
           </div>
-          <div className="flex items-center gap-2 mt-4">1000 Contributors</div>
+          <div className="flex items-center gap-2 mt-4" id="contributorCount">
+            Contributors
+          </div>
         </div>
       </div>
       <div className="absolute right-0 flex flex-col gap-3 py-2 md:items-center md:flex-row md:static">
--- a/components/atoms/ClientOnly/client-only.tsx
+++ b/components/atoms/ClientOnly/client-only.tsx
@@ -1,0 +1,12 @@
+import { useHasMounted } from "lib/hooks/useHasMounted";
+
+function ClientOnly({ children }: { children: React.ReactNode }) {
+  const hasMounted = useHasMounted();
+
+  if (!hasMounted) {
+    return null;
+  }
+  return <>{children}</>;
+}
+
+export default ClientOnly;
--- a/components/organisms/ContributorsTable/contributors-table.tsx
+++ b/components/organisms/ContributorsTable/contributors-table.tsx
@@ -23,11 +23,11 @@ const ContributorTable = ({
 
       {contributors &&
         contributors.length > 0 &&
-        contributors.map((contributor, i) => (
+        contributors.map((contributor) => (
           <ContributorListTableRow
             topic={topic}
             contributor={contributor}
-            key={i}
+            key={contributor.user_id}
             selected={!!selectedContributors?.find((selected) => selected.user_id === contributor.user_id)}
             handleOnSelectContributor={handleSelectContributors}
           />
--- a/helpers/fetchApiData.ts
+++ b/helpers/fetchApiData.ts
@@ -1,14 +1,27 @@
+export function validateListPath(path: string) {
+  const regex =
+    /^lists\/[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}(\/(overview|activity|contributors))?$/;
+
+  return regex.test(path);
+}
+
 export async function fetchApiData<T>({
   path,
   method = "GET",
   headers,
   bearerToken,
+  pathValidator,
 }: {
   path: string;
   method?: "GET" | "POST" | "PUT" | "DELETE" | "PATCH";
   headers?: HeadersInit;
   bearerToken: string;
+  pathValidator(path: string): boolean;
 }) {
+  if (!pathValidator(path)) {
+    return { data: null, error: { status: 400, statusText: "bad request" } };
+  }
+
   const apiUrl = new URL(path, process.env.NEXT_PUBLIC_API_URL);
   const response = await fetch(apiUrl, {
     method,
--- a/helpers/fetchApiData.ts
+++ b/helpers/fetchApiData.ts
@@ -1,7 +1,7 @@
 import { createPagesServerClient } from "@supabase/auth-helpers-nextjs";
 import { GetServerSidePropsContext } from "next";
 
-export async function fetchApiData({
+export async function fetchApiData<T>({
   baseUrl = process.env.NEXT_PUBLIC_API_URL,
   path,
   headers,
@@ -30,7 +30,7 @@ export async function fetchApiData({
     },
   });
 
-  const data = response.ok ? ((await response.json()) as ContributorList) : null;
+  const data = response.ok ? ((await response.json()) as T) : null;
   const { status, statusText } = response;
   const error = response.ok ? null : { status, statusText };
 
--- a/helpers/fetchApiData.ts
+++ b/helpers/fetchApiData.ts
@@ -1,17 +1,16 @@
 export async function fetchApiData<T>({
-  baseUrl = process.env.NEXT_PUBLIC_API_URL,
   path,
   method = "GET",
   headers,
   bearerToken,
 }: {
-  baseUrl?: string;
   path: string;
   method?: "GET" | "POST" | "PUT" | "DELETE" | "PATCH";
   headers?: HeadersInit;
   bearerToken: string;
 }) {
-  const response = await fetch(`${baseUrl}/${path}`, {
+  const apiUrl = new URL(path, process.env.NEXT_PUBLIC_API_URL);
+  const response = await fetch(apiUrl, {
     method,
     headers: {
       ...headers,
--- a/helpers/fetchApiData.ts
+++ b/helpers/fetchApiData.ts
@@ -1,0 +1,38 @@
+import { createPagesServerClient } from "@supabase/auth-helpers-nextjs";
+import { GetServerSidePropsContext } from "next";
+
+export async function fetchApiData({
+  baseUrl = process.env.NEXT_PUBLIC_API_URL,
+  path,
+  headers,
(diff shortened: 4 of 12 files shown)


helpers/fetchApiData.ts
```diff
@@ -0,0 +1,38 @@
+import { createPagesServerClient } from "@supabase/auth-helpers-nextjs";
+import { GetServerSidePropsContext } from "next";
+
+export async function fetchApiData({
```
nickytonline, line 4: I'm wrestling with passing the access token in or abstracting it away into the function. I think it's fine for now, but curious on others thoughts.

layouts/lists.tsx
```diff
@@ -36,21 +39,13 @@ const ListPageLayout = ({ children }: { children: React.ReactNode }) => {
         <div className="info-container container w-full min-h-[100px]">
           <Header>
             {list ? (
-              <ListHeader list={list} listId={list.id} isOwner={isOwner} />
-            ) : isLoading ? (
-              <div className="flex justify-between w-full h-46">
```
nickytonline, line 41: The skeleton wrapper is no longer required as this renders server side via `getServerSideProps`.

pages/lists/[listId]/contributors.tsx
```diff
@@ -1,9 +1,147 @@
...
+  const { listId } = ctx.params as { listId: string };
+  const { data, error: contributorListError } = await fetchApiData({
+    context: ctx,
+    path: `lists/${listId}/contributors`,
+  });
+  const contributors = convertToContributors(data?.data);
+
+  const { data: list, error } = await fetchApiData({ context: ctx, path: `lists/${listId}` });
+
+  return {
```
Comment (nickytonline) [inline pages/lists/[listId]/contributors.tsx:58]: Add sentry here if there are errors?

pages/lists/[listId]/overview.tsx
```diff
@@ -1,7 +1,7 @@
 import ListPageLayout from "layouts/lists";
 
 const ListsOverview = () => {
-  return <div>Lists Overview</div>;
+  return <h2>Lists Overview</h2>;
```
Comment (nickytonline) [inline pages/lists/[listId]/overview.tsx:4]: small a11y win.

next-types.d.ts
```diff
@@ -336,3 +336,12 @@ interface DbUserList {
   readonly created_at: string;
   readonly updated_at: string;
 }
+
+interface DBList {
```
nickytonline, line 340: Is this what the list should be called or do we have another name for it? For example, the user list is `DBUserList`.	It seems to be a generic list for stuff being pulled from `/lists/${id}` in the API.

nickytonline:
@brandonroberts @jpmcb, the `lists/{id}` endpoint doesn't appear to include who created the list. This info is required to determine if the current user can edit the list or not. That info exists when you create a list, so I imagine I can add the property in the API for that endpoint's payload?

brandonroberts:
Yes, that can be added. In the API, each list is tied to the user that created it, so if the logged in user is not the one that created it, an error is returned.

jpmcb:
Yeah shouldn't be too bad! I can grab that if there's no objections (since Brandon has limited availability this week)?

nickytonline:
> Yeah shouldn't be too bad! I can grab that if there's no objections (since Brandon has limited availability this week)?

Thanks John! In the meantime I'll just mock it for adding the edit button.

components/organisms/ContributorsTable/contributors-table.tsx
```diff
@@ -23,11 +23,11 @@ const ContributorTable = ({
 
       {contributors &&
         contributors.length > 0 &&
-        contributors.map((contributor, i) => (
+        contributors.map((contributor) => (
           <ContributorListTableRow
             topic={topic}
             contributor={contributor}
-            key={i}
+            key={contributor.user_id}
```
nickytonline, line 30: This is a guaranteed unique reliable key for the row. If the items were to be reordered, the original index, `i`, would not remain with the same key.

helpers/fetchApiData.ts
```diff
@@ -0,0 +1,29 @@
...
+}) {
+  const apiUrl = new URL(path, process.env.NEXT_PUBLIC_API_URL);
+  const response = await fetch(apiUrl, {
+    method,
+    headers: {
+      ...headers,
+      accept: "application/json",
+      Authorization: `Bearer ${bearerToken}`,
+    },
+  });
```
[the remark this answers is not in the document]
nickytonline: I'm still trying to sort this out which is why I added a validator path, `validateListPath`.

lib/hooks/api/useContributorList.ts
```diff
@@ -0,0 +1,52 @@
...
+  initialData?: {
+    data: DbPRContributor[];
+    meta: Meta;
+  };
+  initialPage?: number;
+  defaultLimit?: number;
+}) => {
+  const [page, setPage] = useState(initialPage);
+  const [limit, setLimit] = useState(defaultLimit);
+  const { data, error, mutate } = useSWR<any>(
```
OgDev-01, line 35: nit: why `any`?

helpers/fetchApiData.ts
```diff
@@ -0,0 +1,42 @@
+export function validateListPath(path: string) {
+  const regex =
+    /^lists\/[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}(\/(overview|activity|contributors)\?limit=\d+)?$/;
+
+  return regex.test(path);
+}
+
+export async function fetchApiData<T>({
```
nickytonline, line 8: A lot of this code is similar to the `publicFetcher`, so maybe that makes more sense to just use that. I had started that prior to knowing that existed.

layouts/lists.tsx
nickytonline: This layout has been reworked to receive props now.

lib/hooks/api/useContributorList.ts
```diff
@@ -0,0 +1,52 @@
...
+    `lists/${listId}/contributors?page=${page}&limit=${limit}`,
+    publicApiFetcher as Fetcher<PagedData<DBListContributor>, Error>,
+    {
+      fallbackData: initialData,
```
nickytonline, line 39: This is the initial data that comes from getServerSideProps.

lib/hooks/useList.ts
```diff
@@ -69,23 +69,8 @@ const useFetchListContributors = (id: string, range = 30) => {
...
-    user: {
-      id: 1,
-    },
-  } as DbUserList;
-  const error = false;
-  const mutate = () => {};
-  // const { data, error, mutate } = useSWR<DbUserList, Error>(
-  //   id ? endpointString : null,
-  //   publicApiFetcher as Fetcher<DbUserList, Error>
-  // );
```
nickytonline, line 88: This was a mocked implementation that was replaced with a working one.

next-types.d.ts
```diff
@@ -342,3 +342,18 @@ interface DbUserSearch {
...
+  id: string;
+  user_id: number;
+  name: string;
+  is_public: boolean;
+  created_at: string;
+  updated_at: string;
+  deleted_at: string | null;
+}
+
+interface PagedData<T> {
```
nickytonline, line 356: This is a new type I've added. We have lots of types for the same thing that only differ on the type of data in the array. We could replace existing types with this generic, but that's for another day.

pages/lists/[listId]/contributors.tsx
Comment (nickytonline) [reply pages/lists/[listId]/contributors.tsx:40]: I'm still not sure of our flow in the app in terms of error reporting.

OgDev-01 (dismissed):
LGTM 🍕

pages/lists/[listId]/contributors.tsx
Comment (brandonroberts) [reply pages/lists/[listId]/contributors.tsx:40]: We install sentry in the `next.config` which sets up error reporting for the app

pages/lists/[listId]/contributors.tsx
```diff
@@ -1,9 +1,110 @@
...
+      pathValidator: validateListPath,
+    }),
+    fetchApiData<DBList>({ path: `lists/${listId}`, bearerToken, pathValidator: validateListPath }),
+  ]);
+
+  if (error?.status === 404) {
+    return {
+      notFound: true,
+    };
+  }
```
Comment (nickytonline) [inline pages/lists/[listId]/contributors.tsx:36]: I'm redirecting to the not found page, but do we have something to handle this already?

lib/hooks/api/useContributorList.ts
nickytonline: I think I might have copy pasted that. Will update.

helpers/fetchApiData.ts
nickytonline: I think I'm just going to go with `publicApiFetcher` that already exists as mentioned elsewhere.

layouts/lists.tsx
brandonroberts: Note: This is a change in how we've done layouts. Layouts are applied here https://github.com/open-sauced/app/blob/beta/pages/_app.tsx#L160, which is why they have only `children` as props.

So each page under `pages/lists` would need to use the layout the same way

pages/lists/[listId]/contributors.tsx
```diff
@@ -1,9 +1,110 @@
...
+  const {
+    isLoading,
+    setPage,
+    data: { data: contributors, meta },
+  } = useContributorsList({ listId: list?.id, initialData });
+
+  return (
+    <ListPageLayout list={list} numberOfContributors={meta.itemCount} isOwner={isOwner}>
+      <div className="container flex flex-col gap-3">
+        <h2>List Contributors</h2>
```
Comment (brandonroberts) [inline pages/lists/[listId]/contributors.tsx:73]: Should have a pagination header at the top also

image

https://github.com/open-sauced/app/blob/beta/components/molecules/TableHeader/table-header.tsx

The search/layout/range is able to be toggled on/off

components/ListHeader/list-header.tsx
```diff
@@ -42,11 +44,13 @@ const ListHeader = ({ list, listId, isOwner }: ListHeaderProps): JSX.Element =>
         <div className="flex flex-col justify-center p-2 header-info grow">
           <div className="flex gap-2">
             <Title level={1} className="!text-2xl font-semibold tracking-tight text-slate-900">
-              {(list && truncateString(list.name, 30)) || "Insights"}
+              {(name && truncateString(name, 30)) || "Insights"}
```
brandonroberts, line 47: nit: (non blocking) just a carry-over
```suggestion
              {(name && truncateString(name, 30)) || "List"}
```

brandonroberts (changes requested):

layouts/lists.tsx
nickytonline: Yeah, I saw that. The overview page doesn't really have anything at the moment and the Activity page does not exist yet.

I made this change so that a layout could benefit from SSR generated props. I'd apply the same pattern to those other pages for lists, or do you think it would cause issues elsewhere? I believe that template is only use on the overview and contributors list atm.

brandonroberts: Yep, this is a new layout/section, so I don't think it will cause issues elsewhere. I'll adjust my overview page after this lands.

pages/lists/[listId]/contributors.tsx
Comment (brandonroberts) [reply pages/lists/[listId]/contributors.tsx:36]: Nothing currently

helpers/fetchApiData.ts
brandonroberts: We can refactor this after if its already working

pages/lists/[listId]/contributors.tsx
Comment (nickytonline) [reply pages/lists/[listId]/contributors.tsx:73]: I added this and now I'm running into Zustand issues with a `Error: Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering.` error. I dug into it a bit and it was apparently fixed in in Zustand 4.0.0-beta3, but we have a newer version and it still occurs. I think I might have to scrap the SSR stuff for now and we revisit it as a whole for the app. 😓

https://github.com/pmndrs/zustand/issues/886

Comment (brandonroberts) [reply pages/lists/[listId]/contributors.tsx:73]: 😭

Comment (nickytonline) [reply pages/lists/[listId]/contributors.tsx:73]: I'll dig a bit more just in case I missed something.

nickytonline:
I saved the markup from view page source (SSR) and compared it to the markup in the devtools after a full client-side render (CSR) and the SSR version is missing usernames and potentially other stuff. I believe this is because there are calls to populate each row on the client side with additional info. I'm guessing potentially some client-side only hooks that aren't kicking in properly during SSR. 🤔 

CleanShot 2023-09-28 at 17 17 45

components/atoms/ClientOnly/client-only.tsx
```diff
@@ -0,0 +1,12 @@
+import { useHasMounted } from "lib/hooks/useHasMounted";
+
+function ClientOnly({ children }: { children: React.ReactNode }) {
```
brandonroberts, line 3: This is useful for other places we've used this pattern also 👍

brandonroberts (approved):
🚢

OgDev-01 (approved):
LGTM... just some minor styles update and we are good

The table should expand to the layouts width(the drawn lines)
explanation

brandonroberts (approved):
Couple minor tweaks that can follow-up

pages/lists/[listId]/overview.tsx
```diff
@@ -12,10 +14,50 @@ import { getAvatarByUsername } from "lib/utils/github";
...
+
+  return {
+    props: {
+      list,
+      numberOfContributors: data?.meta.itemCount || 0,
+    },
+  };
+};
+
+const ListsOverview = ({ list, numberOfContributors }: ListsOverviewProps): JSX.Element => {
```
Comment (nickytonline) [inline pages/lists/[listId]/overview.tsx:60]: @brandonroberts, I've refactored the Overview page so it works with the way the layout currently loads for the Contributors List.

helpers/fetchApiData.ts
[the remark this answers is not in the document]
nickytonline: Looks like my fix worked.

Issue #1812, opened 2023-10-05
Add visual indication that topics can be changed


BekahHW:
I've received feedback a handful of times now that users didn't know they can change the topic.

Currently: 
![image](https://github.com/open-sauced/app/assets/34313413/94f301c2-e04d-4a06-affd-2f12bb3b4bc5)

Proposed change:
![image](https://github.com/open-sauced/app/assets/34313413/8569f3a6-9bb2-4d5f-8aea-7a1776023cf5)

Maybe adding a down triangle selector is a good indication that it can be changed.

Gourav2609:
hello @BekahHW 
i would like to work on it...please assign this to me

bdougie:
Please review the [CONTIBUTING.md](https://github.com/open-sauced/.github/blob/HEAD/CONTRIBUTING.md) for info on how to assign yourself to the issue.

Gourav2609:
thanks @bdougie

Gourav2609:
.take

Gourav2609:
i have a PR regarding this issue.... @bdougie please check PR #1816

Gourav2609:
hello @BekahHW  i have changes according to your review plz review the PR #1825

bdougie:
> hello @BekahHW  i have changes according to your review plz review the PR #1825 

For a timely review, please fill out the pr template screenshots.

Also confirm if you tested your change on mobile and non chromium browsers

Gourav2609:
@bdougie  i have filled the PR template and attached a recording too

Gourav2609:
@bdougie any suggestions on PR #1825

bdougie:
> @bdougie any suggestions on PR #1825

Thanks for the PR. It unfortunately lacks some attention to detail

## The spacing is inconsistent with other elements in the section. Can you make it tighter by removing some padding or margin?

<img width="427" alt="Screen Shot 2023-10-09 at 10 23 45 AM" src="https://github.com/open-sauced/app/assets/5713670/4bcfe695-f497-4c07-903a-e00f3253a33b">

Gourav2609:
@bdougie sure

Gourav2609:
> > @bdougie any suggestions on PR #1825
> 
> Thanks for the PR. It needs some attention to detail
> 
> ## The spacing is inconsistent with other elements in the section. Can you make it tighter by removing some padding or margin?
[...]

PR #1816, opened 2023-10-05
fix: added dropdown icon instead of hash icon

Gourav2609:
## Description

Earlier:

image

Now:
image

this is reference to the Issue #1812 

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [1 ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [ ] 👍 yes
- [1 ] 🙅 no, because they aren't needed

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

No this is not required since i just  replaced a icon

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@
 
 > All notable changes to this project will be documented in this file
 
+## [1.68.0-beta.9](https://github.com/open-sauced/insights/compare/v1.68.0-beta.8...v1.68.0-beta.9) (2023-10-06)
+
+
+### 🍕 Features
+
+* move range filter to list pages header ([#1817](https://github.com/open-sauced/insights/issues/1817)) ([0b82760](https://github.com/open-sauced/insights/commit/0b82760dfd20b14230f8f0b4496b75c3086db139))
+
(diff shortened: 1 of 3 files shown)


OgDev-01:
Thanks @Gourav2609 for turning this around 😊

Gourav2609:
> Thanks @Gourav2609 for turning this around 😊

My pleasure @OgDev-01 😇😇

BekahHW (changes requested):
I don't think we want to change the icon. The hash svg, should contain the hash symbol. If there isn't currently an icon  for the inverted triangle, it should be added and then the line of code that currently uses the hash svg should be updated to the inverted triangle. 

Does that make sense?

nickytonline:
> I don't think we want to change the icon. The hash svg, should contain the hash symbol. If there isn't currently an icon for the inverted triangle, it should be added and then the line of code that currently uses the hash svg should be updated to the inverted triangle.
> 
> Does that make sense?

Just reiterating what @BekahHW mentioned. We have the other icon.

PR #1898, opened 2023-10-15
fix: can't update user profile bio passed 256 chars

### LINKED ISSUE #1446: Bug: can't update user profile bio passed 256 chars
### Describe the bug

We don't see a concise error message about why we can't update the bio profile.

As a UI/UX I want to see something like that:
image

Right now I see only this info
Screenshot 2023-07-29 at 09 07 59

### Steps to reproduce

1. Go to user settings
2. Put 256 char in textarea 
3. Click on the Update Profile

### Browsers

Chrome

_No response_

### Code of Conduct

### Contributing Docs

aryanas159:
## Description
Added state to track the bio, added a div to keep track of bio  length and max length limit on the bio section.

fixes #1446 

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

image

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/components/organisms/UserSettingsPage/user-settings-page.tsx
+++ b/components/organisms/UserSettingsPage/user-settings-page.tsx
@@ -51,6 +51,7 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
   const [timezone, setTimezone] = useState("");
   const [userInfo, setUserInfo] = useState<DbUser>();
   const [email, setEmail] = useState<string | undefined>("");
+  const [bio, setBio] = useState<string | undefined>("");
   const [emailPreference, setEmailPreference] = useState<EmailPreferenceType>({
     // eslint-disable-next-line camelcase
     display_email: false,
@@ -71,7 +72,8 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
       setEmail(response.email);
       setDisplayLocalTime(response.display_local_time);
       setCoupon(response.coupon_code);
-      formRef.current!.bio.value = response.bio;
+      //formRef.current!.bio.value = response.bio;
+      setBio(response.bio);
       formRef.current!.url.value = response.url;
       formRef.current!.twitter_username.value = response.twitter_username;
       formRef.current!.company.value = response.company;
@@ -162,7 +164,7 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
     const payload: UpdateUserPayload = {
       name: formRef.current!.nameInput.value,
       email,
-      bio: formRef.current!.bio.value,
+      bio: bio,
       // eslint-disable-next-line camelcase
       twitter_username: formRef.current!.twitter_username.value,
       company: formRef.current!.company.value,
@@ -223,11 +225,15 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
             <div className="flex flex-col gap-2">
               <label className="text-sm font-normal text-light-slate-11">Bio</label>
               <textarea
+                maxLength={255}
                 rows={4}
                 placeholder="Tell us about yourself."
                 className="px-3 py-2 rounded-lg bg-light-slate-4 disabled:cursor-not-allowed "
                 name="bio"
+                value={bio}
+                onChange={(e) => setBio(e.target.value)}
               ></textarea>
+              <div>{bio?.length}/255</div>
             </div>
             <TextInput
               className="font-medium bg-light-slate-4 text-light-slate-11"


components/organisms/UserSettingsPage/user-settings-page.tsx
```diff
@@ -223,11 +225,15 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
...
               <textarea
+                maxLength={255}
                 rows={4}
                 placeholder="Tell us about yourself."
                 className="px-3 py-2 rounded-lg bg-light-slate-4 disabled:cursor-not-allowed "
                 name="bio"
+                value={bio}
+                onChange={(e) => setBio(e.target.value)}
               ></textarea>
+              <div>{bio?.length}/255</div>
```
OgDev-01, line 236: nit(blocking): reduce font size a little and use `p` tag instead of `div`
```suggestion
              <p className="text-sm">{bio?.length}/255</p>
```

OgDev-01 (changes requested):

components/organisms/UserSettingsPage/user-settings-page.tsx
```diff
@@ -71,7 +72,8 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
       setEmail(response.email);
       setDisplayLocalTime(response.display_local_time);
       setCoupon(response.coupon_code);
-      formRef.current!.bio.value = response.bio;
+      //formRef.current!.bio.value = response.bio;
```
OgDev-01, line 75: nit: remove dead code

components/organisms/UserSettingsPage/user-settings-page.tsx
```diff
@@ -162,7 +164,7 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
     const payload: UpdateUserPayload = {
       name: formRef.current!.nameInput.value,
       email,
-      bio: formRef.current!.bio.value,
+      bio: bio,
```
nickytonline, line 167: refactor (blocking): use object property shorthand.

```suggestion
      bio,
```

components/organisms/UserSettingsPage/user-settings-page.tsx
```diff
@@ -51,6 +51,7 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
   const [timezone, setTimezone] = useState("");
   const [userInfo, setUserInfo] = useState<DbUser>();
   const [email, setEmail] = useState<string | undefined>("");
+  const [bio, setBio] = useState<string | undefined>("");
```
nickytonline, line 54: The initial value is set to an empty string. If the value is never set to `undefined` let the type be inferred. If it does get set to undefined, do not set an initial value, i.e. set to `undefined` by default?

```suggestion
  const [bio, setBio] = useState("");
```

nickytonline (changes requested):

components/organisms/UserSettingsPage/user-settings-page.tsx
aryanas159: image
After changes.
I changed text size to xs since sm didn't make any difference in the text size.

OgDev-01 (approved):
LG 🍕

nickytonline:
The current fix prevents you from adding more text in the bio via `max-length` on the input, but the initial request in the #1446 was suggesting a message indicating the bio is too long. I agree that it'd be good to have a message even if we prevent them on the client-side from surpassing the allowed length.

image

aryanas159:
> The current fix prevents you from adding more text in the bio via `max-length` on the input, but the initial request in the #1446 was suggesting a message indicating the bio is too long. I agree that it'd be good to have a message even if we prevent them on the client-side from surpassing the allowed length.
> 
> image

I have added a ternary operation that shows the "bio too long" message once we exceed the limit.
image

components/organisms/UserSettingsPage/user-settings-page.tsx
```diff
@@ -227,7 +228,14 @@ const UserSettingsPage = ({ user }: userSettingsPageProps) => {
...
                 className="px-3 py-2 rounded-lg bg-light-slate-4 disabled:cursor-not-allowed "
                 name="bio"
+                value={bio}
+                onChange={(e) => setBio(e.target.value)}
               ></textarea>
+              {bio?.length > 255 ? (
+                <p className="text-light-red-10 text-xs">Bio too long</p>
+              ) : (
+                <p className="text-xs">{bio?.length}/255</p>
+              )}
```
nickytonline, line 238: refactor (blocking): `aria-live="assertive"` should be used for the validation error and `"polite"` for the character count. See https://www.smashingmagazine.com/2023/02/guide-accessible-form-validation/#instant-validation and https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

```suggestion
              {bio?.length > 255 ? (
                <p aria-live="assertive" className="text-light-red-10 text-xs">Bio too long</p>
              ) : (
                <p aria-live="polite" className="text-xs">{bio?.length}/255</p>
              )}
```

nickytonline (changes requested):
This is looking good, just one more change before we can merge this.

nickytonline (approved):
Thanks for making the changes @aryanas159 and for making the application more accessible.

:shipit:

PR #2061, opened 2023-11-06
feat: make single select component searchable

### LINKED ISSUE #1445: Feature: Support typing characters when timezone dropdown is opened
### Type of feature

🍕 Feature

### Current behavior

Currently, there's no ability to type characters to focus on the first matched option when the dropdown is opened:

Screenshot from 2023-07-28 18-01-00

### Suggested solution

There should be support for this type of keyboard interaction. The Radix select component currently only supports a limited set of keys (https://www.radix-ui.com/docs/primitives/components/select#keyboard-interactions). With this feature, users would be able to open a select dropdown and start typing characters, and the dropdown would attempt to find and focus on the first option that matches the typed characters. This functionality would enable users to quickly navigate to the desired option by typing part of its label or value.

If this issue gets prioritized and it makes sense to be resolved, I would like to try providing this fix.

### Additional context

_No response_

### Code of Conduct

### Contributing Docs

OgDev-01:
## Description
This PR adds search support to the single select component

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

closes #1445 

https://github.com/open-sauced/app/assets/62995161/59916a94-8f22-43c8-8a88-280768ff6436

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook


brandonroberts:
I think the search option should be configurable depending on where its used

OgDev-01:
> I think the search option should be configurable depending on where its used

I've made it optional @brandonroberts... you can take a second peak when you have the time 👍

components/atoms/Select/single-select.tsx
```diff
@@ -27,12 +40,41 @@ const SingleSelect = ({ placeholder, value, onValueChange, options, position }:
...
+              onValueChange={setInputValue}
+            />
+            <CommandGroup className="flex flex-col !px-0 overflow-scroll max-h-[25rem]">
+              {options.map((option, index) => (
+                <CommandItem
+                  key={`timezone_${index}`}
+                  onMouseDown={(e) => {
+                    e.preventDefault();
+                    e.stopPropagation();
+                  }}
```
nickytonline, line 59: I did a test and this isn't required.

nickytonline (changes requested):
Making it searchable breaks keyboard navigation for the component when `isSearchable` is true.

When `isSearchable` is enabled, you can navigate the list, but can no longer press <kbd>ENTER</kbd> to select an item in the list.

CleanShot 2023-11-08 at 13 26 19

With `isSearchable` equal to false, the `<SingleSelect />` still works with the <kbd>ENTER</kbd> key for selecting an item in the list.

CleanShot 2023-11-08 at 13 24 32

The problem is the search field keeps the focus disallowing the items to be selected with the keyboard.

NsdHSO (approved):

OgDev-01:
> I did a test and this isn't required. 

Talking it off prevents the enter key from working with the current fix i have

OgDev-01:
One thing we might need to also consider is clearing out the selected value... This is something i'll push in on a separate PR

brandonroberts (approved):

nickytonline (commented):
There's still some issues in regards to focus. when the select opens, there is no focus on the search. If you start navigating with the up and down arrows, you can't leave the list without pressing enter and you're unable to get focus on the search field in the list. I'll drop a video shortly.

nickytonline:
Here's as short video demonstrating the issue @OgDev-01. https://www.loom.com/share/85ec68a1085e4484b947b2a93cd4e023

nickytonline:
We can go through this on Thursday during our pairing session @OgDev-01.= if you're still blocked.

OgDev-01:
Hey @nickytonline, i'm able to add the focus state to the select but still can't find a walk around to making the input focus on   select open or using the direction navigation. We can revisit this another time as it is currently not a blocker.

components/atoms/Select/single-select.tsx
```diff
@@ -9,15 +10,50 @@ interface SingleSelectProps {
...
+    }
+  }, [isOpen, isSearchable]);
+
   return (
-    <Select value={value} onValueChange={(value) => onValueChange(value)}>
+    <Select
+      value={value}
+      onValueChange={(value) => {
+        onValueChange(value);
+      }}
```
nickytonline, line 47: refactor (blocking): just set the prop to the function prop passed in.

```suggestion
      onValueChange={onValueChange}
```

components/atoms/Select/single-select.tsx
```diff
@@ -27,12 +63,45 @@ const SingleSelect = ({ placeholder, value, onValueChange, options, position }:
         <SelectValue className="w-10 truncate" placeholder={placeholder ?? "Select time zone"} />
       </SelectTrigger>
 
-      <SelectContent position={position ?? "item-aligned"} className="z-50 bg-white">
-        {options.map((option, index) => (
-          <SelectItem title={option.label} key={`timezone_${index}`} value={option.value}>
-            {truncateString(option.label, 30)}
-          </SelectItem>
-        ))}
+      <SelectContent position={position ?? "item-aligned"} className="z-50 relative bg-white">
```
nickytonline, line 66: Curious if the z-index is necessary. I know it was here previously.

components/atoms/Select/single-select.tsx
```diff
@@ -9,15 +10,50 @@ interface SingleSelectProps {
...
+  inputPlaceholder,
+  isSearchable = false,
+}: SingleSelectProps) => {
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [inputValue, setInputValue] = useState("");
+  const [isOpen, setIsOpen] = useState(false);
+
+  const filteredOptions = options.filter((option) => option.label.toLowerCase().includes(inputValue.toLowerCase()));
+
+  const getOptions = () => (isSearchable ? filteredOptions : options);
```
nickytonline, line 34: refactor (blocking): this doesn't need to be a function. The variable could be called `items`, or something similar and used in the rest of the file instead of calling the function everytime.

```suggestion
  const items = isSearchable ? filteredOptions : options;
```

nickytonline (changes requested):
A couple changes and this is good to go @OgDev-01.

nickytonline:
Once #2340 is merged @OgDev-01, you can add this story for this PR to it.

```
export const Searchable: Story = {
  args: { ...baseProps, isSearchable: true },
};
```

components/atoms/Select/single-select.tsx
OgDev-01: It's needed to keep the items above other elements when open

OgDev-01:
I've included the story test for searchable selects cc: @nickytonline 
image

nickytonline (approved):
Thanks @OgDev-01! :shipit:

Issue #2051, opened 2023-11-05
Feature: populate new high light URL.


a0m0rajab:
### Type of feature

🍕 Feature

### Current behavior

The new highlight button in the AI extension is leading to new page without populating the URL, it would be great to get the URL populated. Related to: https://github.com/open-sauced/ai/issues/256

### Suggested solution

We need to check if it's a new highlight and then check the URL to see if it's available to populate it: 

related code: 

https://github.com/open-sauced/app/blob/cc93f2c26cbd7812557407f569b62ce575ab937e/pages/feed/index.tsx#L101-L108

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [ ] I agree to follow this project's Contribution Docs

a0m0rajab:
.take

PR #2062, opened 2023-11-07
feat: add params handling for new PR

a0m0rajab:
## Description

This PR add handling PR Url in the params, 

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Closes https://github.com/open-sauced/app/issues/2051

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook


components/molecules/HighlightInput/highlight-input-form.tsx
```diff
@@ -500,6 +500,13 @@ const HighlightInputForm = ({ refreshCallback }: HighlightInputFormProps): JSX.E
   }, 250);
 
   useEffect(() => {
+    const queryParams = new URLSearchParams(window.location.search);
```
nickytonline, line 503: Refactor (blocking): Instead of `new URLSearchParams(window.location.search)` prefer the `useRouter` hook.

e.g.

```
  const router = useRouter();
  const { prurl } = router.query;
```

Also, this doesn't need to be in the a `useEffect`.

pages/feed/index.tsx
```diff
@@ -101,10 +101,15 @@ const Feeds: WithPageLayout<HighlightSSRProps> = (props: HighlightSSRProps) => {
...
+    const prURL = queryParams.get("prurl");
     const signInRequired = queryParams.get("signIn");
 
     if (newHighlight && signInRequired) {
-      signIn({ provider: "github", options: { redirectTo: `${window.location.origin}/feed?new=${newHighlight}` } });
+      let query = "";
+      for (const [key, value] of queryParams.entries()) {
+        query += `${key}=${value}&`;
+      }
+      signIn({ provider: "github", options: { redirectTo: `${window.location.origin}/feed?${query}` } });
```
nickytonline, line 112: Refactor (blocking): Use `queryParams` built in `.toString()` which gets used implicitly when used in a template string. Another benefit to using it is it encodes the querystring values.

```suggestion
      signIn({ provider: "github", options: { redirectTo: `${window.location.origin}/feed?${queryParams}` } });
```

nickytonline (changes requested):
Just a couple chang requests @a0m0rajab.

components/molecules/HighlightInput/highlight-input-form.tsx
a0m0rajab: Thanks for the suggestion, I would love to get help with this. 

- Out of curiosity: why would you suggest the useRouter? 

When I am using the useRouter outside the useEffect it gives me too many rendering error; how could I solve that? 

```
Error: Too many re-renders. React limits the number of renders to prevent an infinite loop.
```
here is the code: 
```js
const router = useRouter();
const { prurl } = router.query;
if (prurl) {
  setHighlightLink(prurl as string);
}
```

a0m0rajab: I could get it to work by using this:

```js
  const router = useRouter();
  const { prurl } = router.query;
  useEffect(() => {
    if(!prurl) return;
    setHighlightLink(prurl as string);
  }, [prurl]);
```

OgDev-01:
Great work @a0m0rajab 👏👏. Would you mind uploading a screen record of the behaviour now and before to enable a quicker review process?

components/molecules/HighlightInput/highlight-input-form.tsx
nickytonline: Yeah, that's my bad. I think it needs to remain in a `useEffect` like you have. I suggested the `useRouter` as we're using pretty much everywhere for stuff like this.

a0m0rajab: I have worked on this for almost an hour yesterday lol. but it was worth it, my previous logic were broken to be honest. and I think the userRouter is more suited right now.

pages/feed/index.tsx
```diff
@@ -101,10 +101,11 @@ const Feeds: WithPageLayout<HighlightSSRProps> = (props: HighlightSSRProps) => {
   useEffect(() => {
     const queryParams = new URLSearchParams(window.location.search);
     const newHighlight = queryParams.get("new");
+    const prURL = queryParams.get("prurl");
```
nickytonline, line 104: This appears to not be used anywhere? If so, it can be removed.

OgDev-01 (approved):
LGTM 🍕

a0m0rajab:
@OgDev-01 thanks for that. Here is the before and after videos: 

# Before 
https://github.com/open-sauced/app/assets/18273833/8f50d396-73e9-4be3-8c9f-4e7439143b47

# After

https://github.com/open-sauced/app/assets/18273833/312b5fcc-c1f5-43d2-8e0b-712968618e2d

pages/feed/index.tsx
a0m0rajab: Thanks, just removed it. not sure how I did not notice that lol

brandonroberts (approved):

PR #2118, opened 2023-11-12
fix: pr overviews in repo table now accurate

### LINKED ISSUE #2097: Bug: PR overview not adding up
### Describe the bug

In some cases, PR overview percentage numbers do not add up to 100%.

### Steps to reproduce

For OpenSauced docs on this [insight page](https://app.opensauced.pizza/pages/BekahHW/230/repositories), it adds up to 92%. Visually, you can see that the line doesn't extend as far as the repo above. 

92% of 12 is 11. We currently have one in draft, so I would guess that for some reason, the gray part of the line isn't rendering.  

You'll notice from the full image that the gray line _does_ render on intro, though. 

image

image

image

### Browsers

_No response_

_No response_

### Code of Conduct

### Contributing Docs

OgDev-01:
## Description
This PR fixes the repo table pr velocity chart display i inaccuracy issue

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

fixes #2097 

### Before
image

### After
image

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/components/atoms/PullRequestOverviewChart/pull-request-overview-chart.tsx
+++ b/components/atoms/PullRequestOverviewChart/pull-request-overview-chart.tsx
@@ -10,7 +10,7 @@ const PullRequestOverviewChartBar: React.FC<PullRequestOverviewChartBarProps & R
   type,
   ...event
 }) => {
-  const getBarColor = (type: string) => {
+  const getBarColor = (type: string | undefined) => {
     switch (type) {
       case "open":
         return "bg-light-grass-9";
@@ -29,7 +29,7 @@ const PullRequestOverviewChartBar: React.FC<PullRequestOverviewChartBarProps & R
     <div
       {...event}
       className={`
-        ${getBarColor(type!)}
+        ${getBarColor(type)}
         ${percent === 0 ? "hidden" : "block"}
         transition-all shrink-0 duration-500 ease-in-out rounded-full`}
       style={{ width: `${percent}%` }}
--- a/components/atoms/PullRequestOverviewChart/pull-request-overview-chart.tsx
+++ b/components/atoms/PullRequestOverviewChart/pull-request-overview-chart.tsx
@@ -10,24 +10,29 @@ const PullRequestOverviewChartBar: React.FC<PullRequestOverviewChartBarProps & R
   type,
   ...event
 }) => {
+  const getBarColor = (type: string) => {
+    switch (type) {
+      case "open":
+        return "bg-light-grass-9";
+      case "merged":
+        return "bg-purple-600";
+      case "closed":
+        return "bg-light-red-9";
+      case "draft":
+        return "bg-light-slate-9";
+      default:
+        return "";
+    }
+  };
+
   return (
     <div
       {...event}
       className={`
-        ${
-          type === "open"
-            ? "bg-light-grass-9"
-            : type === "merged"
-            ? "bg-purple-600"
-            : type === "closed"
-            ? "bg-light-red-9"
-            : type === "draft"
-            ? "bg-light-slate-9"
-            : ""
-        }
-    ${percent === 0 && "hidden"}
+        ${getBarColor(type!)}
+        ${percent === 0 ? "hidden" : "block"}
         transition-all shrink-0 duration-500 ease-in-out rounded-full`}
-      style={{ width: `${percent}%` }}
+      style={{ width: `${Math.round(percent!)}%` }}
     ></div>
   );
 };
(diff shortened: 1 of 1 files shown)


nickytonline:
There's a before image. Do you have an after image to compare @OgDev-01?

components/atoms/PullRequestOverviewChart/pull-request-overview-chart.tsx
```diff
@@ -10,24 +10,29 @@ const PullRequestOverviewChartBar: React.FC<PullRequestOverviewChartBarProps & R
...
-            : type === "merged"
-            ? "bg-purple-600"
-            : type === "closed"
-            ? "bg-light-red-9"
-            : type === "draft"
-            ? "bg-light-slate-9"
-            : ""
-        }
-    ${percent === 0 && "hidden"}
+        ${getBarColor(type!)}
```
nickytonline, line 32: Refactor (blocking): The `!` is not necessary. Avoid changing the type whenever possible. It looks like in the old code, it supported `undefined` as it returned `""` which is the default for the switch case now.

`const getBarColor = (type: string|undefined) => {...}` should do the trick.

components/atoms/PullRequestOverviewChart/pull-request-overview-chart.tsx
```diff
@@ -10,24 +10,29 @@ const PullRequestOverviewChartBar: React.FC<PullRequestOverviewChartBarProps & R
...
+        ${percent === 0 ? "hidden" : "block"}
         transition-all shrink-0 duration-500 ease-in-out rounded-full`}
-      style={{ width: `${percent}%` }}
+      style={{ width: `${Math.round(percent!)}%` }}
```
nickytonline, line 35: This appears to be the actual fix?

nickytonline (changes requested):

components/atoms/PullRequestOverviewChart/pull-request-overview-chart.tsx
```diff
@@ -58,40 +63,40 @@ const PullRequestOverviewChart: React.FC<PullRequestOverviewChartProps> = ({
...
       {/* Merged */}
-      {merged && merged > 0 && (
+      {merged && merged > 0 ? (
         <PullRequestOverviewChartBar
           onMouseOver={() => setOverviewDetails({ type: "merged", percent: merged })}
           percent={getPercentage(merged)}
           type="merged"
         />
-      )}
+      ) : null}
```
OgDev-01, line 81: This was the actual fix... for some reasons, the `&&` condition renders a `"0"` in the dom for all `0%` type.

components/atoms/PullRequestOverviewChart/pull-request-overview-chart.tsx
OgDev-01: This is just a small improvement to remove decimal points from the percent in the markup styles

jpmcb: If the intent is to have all lines add up to 100%, there are edge cases here that won't be accounted for. For example:

```js
console.log(
    Math.round(50.01),
    Math.round(49.99 / 3),
    Math.round(49.99 / 3),
    Math.round(49.99 / 3)
);
// 50 17 17 17
```

which adds up to 101.

Also, this will drop any precision on 0 below `0.5` which will cause those parts to not render:

```js
console.log(Math.round(0.4));
// 0
```

---

We may want to consider adding additional precision onto these calculations if we care about having all parts render (unless it is truly 0) and we want it to all add up to 100

jpmcb: > This is just a small improvement to remove decimal points from the percent in the markup styles

Makes sense. And if we're going for more or less "close" to adding up to 100 and we're ok dropping percents below `0.4`, then we'll just want to be aware of these edge cases or account for them if that's unacceptable from the design perspective.

components/atoms/PullRequestOverviewChart/pull-request-overview-chart.tsx
nickytonline: It's because `draft` is `0` which is falsy, so it stops evaluating the rest and you end up with a `0`.

components/atoms/PullRequestOverviewChart/pull-request-overview-chart.tsx
nickytonline: nit: Since this is not part of the fix, consider this for another PR as it's not related to the fix.

nickytonline (approved):
Left some comments, but this is good to ship. :shipit:

bdougie (approved):
Tested and looks good in app

components/atoms/PullRequestOverviewChart/pull-request-overview-chart.tsx
OgDev-01: reverted the change for now since the previous one broke nothing so we can move on. will revisit another time .

Thanks for the observations 👍

PR #2153, opened 2023-11-15
refactor: move limit query from state to urlParams

### LINKED ISSUE #2154: refactor: move limit query from `useState` to `urlSearchParams` across application
### Type of feature

🧑‍💻 Refactor

### Current behavior

Similar to what i started in https://github.com/open-sauced/app/pull/2047, we need to also store the limit filter state in the browsers url query. This is also related to https://github.com/open-sauced/app/issues/712.

### Suggested solution

refactor the useState approach to using the browsers urlQuery params

### Additional context

_No response_

### Code of Conduct

### Contributing Docs

OgDev-01:
## Description
This PR refactors how the limit query state is stored across the application. Now we have the limit in the url of the pages

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

closes #2154

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook


lib/hooks/api/useContributorList.ts
```diff
@@ -30,16 +30,15 @@ export const useContributorsList = ({
     meta: Meta;
   };
   initialPage?: number;
-  defaultLimit?: number;
+  defaultLimit?: string;
   defaultRange?: string;
 }) => {
-  const [range, setRange] = useState(defaultRange); // [start, end
```
nickytonline, line 36: Refactor (nit): Prefer a separate pull request for the changes for range as it is not related to the current refactor.

lib/hooks/api/useContributorList.ts
```diff
@@ -30,16 +30,15 @@ export const useContributorsList = ({
...
   const [page, setPage] = useState(initialPage);
-  const [limit, setLimit] = useState(defaultLimit);
+
   const query = new URLSearchParams();
   query.append("page", page.toString());
-  query.append("limit", limit.toString());
-  query.append("range", range);
+  query.append("limit", defaultLimit ?? "10");
```
nickytonline, line 40: Refactor (blocking): The `defaultLimit` already has a default value of `"10"`, so `?? "10"` is note required.

```suggestion
  query.append("limit", defaultLimit);
```

pages/lists/[listId]/activity.tsx
```diff
@@ -31,16 +31,23 @@ interface ContributorListPageProps {
   isOwner: boolean;
 }
 
+export type ServerFilterParams = {
```
Comment (nickytonline) [inline pages/lists/[listId]/activity.tsx:34]: Naming (nit): No need to specify `Server` in the type name.

```suggestion
export type FilterParams = {
```

Also, `FilterParams` is used on the list activity page but also elsewhere. Consider moving the type elsewhere and import it into the list activity page.

nickytonline (changes requested):

pages/lists/[listId]/activity.tsx
```diff
@@ -31,16 +31,23 @@ interface ContributorListPageProps {
...
+
 export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
   const supabase = createPagesServerClient(ctx);
 
   const {
     data: { session },
   } = await supabase.auth.getSession();
   const bearerToken = session ? session.access_token : "";
-  const { listId, range: rawRange } = ctx.params as { listId: string; range: string };
+  const { listId, range: rawRange, limit: RawLimit } = ctx.params as ServerFilterParams;
```
Comment (nickytonline) [inline pages/lists/[listId]/activity.tsx:47]: Refactor (blocking): `RawLimit` should be camel case as it's a JavaScript variable and not a type.

```suggestion
  const { listId, range: rawRange, limit: rawLimit } = ctx.params as ServerFilterParams;
```

nickytonline (changes requested):
This is looking good @OgDev-01, just some change requests.

brandonroberts:
@OgDev-01 I think the changes are good, but using strings as something that should be a number seems off. I know that's how they are in the URL, but I think we should convert them to a numbers at some point.

nickytonline:
> @OgDev-01 I think the changes are good, but using strings as something that should be a number seems off. I know that's how they are in the URL, but I think we should convert them to a numbers at some point.

Yeah it seemed off to me too. I mentioned it in the changes for the range changes PR as well although it wasn't blocking the PR. Agree with Brandon here even if they become strings in the end.

We can revisit it for `range` after.

components/atoms/Select/limit-select.tsx
```diff
@@ -6,13 +6,17 @@ import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from "c
...
+  defaultValue?: limitSelectMap;
   onChange?: (value: string) => void;
   options: { name: string; value: string }[];
   className?: string;
 }
-const LimitSelect = ({ onChange, options, className, placeholder }: LimitSelectProps) => {
-  const [selected, setSelected] = useState<null | string>(null);
-  const handleSelected = (value: string) => {
+
+export type limitSelectMap = "10" | "20" | "30" | "40" | "50";
```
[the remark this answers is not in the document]
jpmcb: @nickytonline there are a few places within the API where we default to some set limit. Example:

https://github.com/open-sauced/api/blob/24892059033ee8e9855517d887475cdb54f8681f/src/user/dtos/top-users.dto.ts#L25-L36

But as far as I can tell, these numbers are arbitrary. The upper limit on many of these are intended to prevent full table scans.

pages/lists/[listId]/contributors.tsx
```diff
@@ -69,10 +70,13 @@ const ContributorsListPage = ({ list, initialData, isError, isOwner }: Contribut
...
     setPage,
-    setLimit,
-    setRange,
     data: { data: contributors, meta },
-  } = useContributorsList({ listId: list?.id, initialData, defaultRange: range as string });
+  } = useContributorsList({
+    listId: list?.id,
+    initialData,
+    defaultRange: range as string,
+    defaultLimit: limit as string,
```
Comment (jpmcb) [inline pages/lists/[listId]/contributors.tsx:78]: Looks like do a cast here we cast this number to a string to be consumed in the URL? Like brandon mentioned, I'm also curious if there's a way to prevent storing strings as numbers (especially if it'll be converted to a number eventually and cast back to a string here)

Comment (nickytonline) [reply pages/lists/[listId]/contributors.tsx:78]: This one is king of inevitable because it comes from the router via the querystring. It's type is `string | string[] | undefined`, so in this case, this would need to become most likely this

`defaultLimit: Number(limit as string),`

OgDev-01:
Hey @open-sauced/engineering, I've reverted the limit type to number and converted it to string in the appropriate areas. this is ready for a second review 🍕

jpmcb (approved):

brandonroberts (approved):

PR #2180, opened 2023-11-17
feat: implemented feature flag support

### LINKED ISSUE #2179: Feature: Implement Feature Flag Support in the App
Use PostHog to leverage their feature flagging. Implement a function to get all feature flags for a given user.

nickytonline:
## Description

Our initial offering of feature flagging for the application. This will be used initially to feature flag a graph in the list activity page that will be in a follow up PR.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Closes #2179

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

I've created the feature flag `contributions_evolution_by_type` in prod and beta environment, but it'd be good to just make sure everything there is lined up correctly before going to prod.

### DIFF
--- a/components/shared/feature-flagged.ts
+++ b/components/shared/feature-flagged.ts
@@ -1,0 +1,14 @@
+import { FeatureFlag } from "lib/utils/server/feature-flags";
+
+interface FeatureFlaggedProps {
+  flag: FeatureFlag;
+  children: React.ReactElement;
+  featureFlags: Record<FeatureFlag, boolean>;
+}
+export const FeatureFlagged = ({ flag, featureFlags, children }: FeatureFlaggedProps) => {
+  if (featureFlags[flag]) {
+    return children;
+  }
+
+  return null;
+};
--- a/lib/utils/server/feature-flags.ts
+++ b/lib/utils/server/feature-flags.ts
@@ -1,0 +1,17 @@
+import { PostHog } from "posthog-node";
+
+export type FeatureFlag = "contributions_evolution_by_type";
+
+export async function getAllFeatureFlags(userId: number) {
+  const client = new PostHog(
+    process.env.NEXT_PUBLIC_POSTHOG_ID as string,
+
+    { host: "https://app.posthog.com" }
+  );
+
+  const featureFlags = await client.getAllFlags(`${userId}`);
+
+  await client.shutdownAsync();
+
+  return featureFlags;
+}
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
     "octokit": "^2.0.14",
     "path-to-regexp": "^6.2.1",
     "posthog-js": "^1.83.0",
+    "posthog-node": "^3.1.3",
     "react": "^18.2.0",
     "react-csv": "^2.2.2",
     "react-day-picker": "^8.7.1",
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -56,6 +56,7 @@
         "octokit": "^2.0.14",
         "path-to-regexp": "^6.2.1",
         "posthog-js": "^1.83.0",
+        "posthog-node": "^3.1.3",
         "react": "^18.2.0",
         "react-csv": "^2.2.2",
         "react-day-picker": "^8.7.1",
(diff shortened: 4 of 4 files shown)


components/shared/feature-flagged.ts
```diff
@@ -0,0 +1,14 @@
+import { FeatureFlag } from "lib/utils/server/feature-flags";
+
+interface FeatureFlaggedProps {
+  flag: FeatureFlag;
+  children: React.ReactElement;
+  featureFlags: Record<FeatureFlag, boolean>;
+}
+export const FeatureFlagged = ({ flag, featureFlags, children }: FeatureFlaggedProps) => {
```
nickytonline, line 8: A wrapper component to render a component or not that works server and client side.

lib/utils/server/feature-flags.ts
```diff
@@ -0,0 +1,17 @@
+import { PostHog } from "posthog-node";
+
+export type FeatureFlag = "contributions_evolution_by_type";
```
nickytonline, line 3: This union will expand over time.

lib/utils/server/feature-flags.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export async function getAllFeatureFlags(userId: number) {
+  const client = new PostHog(
```
nickytonline, line 6: Based off of https://posthog.com/docs/libraries/next-js

components/shared/feature-flagged.ts
nickytonline: This could also be done with e.g.

`{ featureFlags["contributions_evolution_by_type"] && <Component /> }`

but it feels more natural to use JSX in JSXland.

package.json
```diff
@@ -92,6 +92,7 @@
     "octokit": "^2.0.14",
     "path-to-regexp": "^6.2.1",
     "posthog-js": "^1.83.0",
+    "posthog-node": "^3.1.3",
```
nickytonline, line 95: Note that we're using feature flag support server-side because some browsers and/or browser extensions can block PostHog.

image

components/shared/feature-flagged.ts
bdougie: I defer to engineers, but this is good to know.

bdougie (approved):

components/shared/feature-flagged.ts
nickytonline: I guess it's a preference as well, so if folks don't like it, we can do the alternate way I mentioned. Not a hill I want to die on. 😅

lib/utils/server/feature-flags.ts
brandonroberts: Should PostHog be instantiated on each function call?

nickytonline: The example in their docs that in linked to send to indicate yes, e.g.

```

export async function getServerSideProps(ctx) {

  const session = await getServerSession(ctx.req, ctx.res)
  let flags = null

  if (session) {
    const client = new PostHog(
      process.env.NEXT_PUBLIC_POSTHOG_KEY,
      {
        host: process.env.NEXT_PUBLIC_POSTHOG_HOST,
      }
    )

    flags = await client.getAllFlags(session.user.email);
    client.capture(session.user.email, 'loaded blog article', { url: ctx.req.url })

    await client.shutdownAsync()
  }
  ...
}
```

components/shared/feature-flagged.ts
brandonroberts: I don't have a strong opinion on it, but I agree with using JSX in JSXland 😅

brandonroberts (approved):

components/shared/feature-flagged.ts
jpmcb: This works for me!

PR #2168, opened 2023-11-17
feat: Svelte added to interests dropdown list

brittneypostma:
## Description

added svelte to interests dropdown list

## What type of PR is this? (check all applicable)

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Tickets & Documents

Fixes #2167

## Added tests?

- [ ] 👍 yes
- [x] 🙅 no, because they aren't needed
- [ ] 🙋 no, because I need help

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook
- [ ] 🙅 no documentation needed

PR #2192, opened 2023-11-21
fix: improved type inference of interests and recommendations

nickytonline:
## Description

I discovered we had an `interestType` inferred type of the interests array and noticed we weren't leveraging it elsewhere and instead we were explicitly typing it elsewhere. A bug was found in #2196 that would have been caught by the TypeScript compiler if we had used the `interestType` everywhere instead of explicitly typing the type elsewhere as it got out of sync with the `interestType`.

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- More context available from my blog post about this PR, [TypeScript: Infer Types to Avoid Explicit Types](https://dev.to/nickytonline/infer-types-to-avoid-explicit-types-1pnn)
- Relates to #2168 and also #2196

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/components/atoms/LanguagePill/LanguagePill.tsx
+++ b/components/atoms/LanguagePill/LanguagePill.tsx
@@ -21,35 +21,17 @@ import HacktoberfestIcon from "img/icons/interests/hacktoberfest.svg";
 import CloJureIcon from "img/icons/interests/clojure.svg";
 
 import topicNameFormatting from "lib/utils/topic-name-formatting";
+import { InterestType } from "lib/utils/getInterestOptions";
 
 interface LanguagePillProps {
-  topic:
-    | "react"
-    | "javascript"
-    | "python"
-    | "ML"
-    | "AI"
-    | "rust"
-    | "ruby"
-    | "c"
-    | "cpp"
-    | "csharp"
-    | "php"
-    | "java"
-    | "svelte"
-    | "typescript"
-    | "golang"
-    | "vue"
-    | "Kubernetes"
-    | "hacktoberfest"
-    | "clojure"
-    | string;
+  topic: InterestType;
   classNames?: string;
   onClick?: () => void;
 }
+
 const LanguagePill = ({ topic, classNames, onClick }: LanguagePillProps) => {
-  const renderTopicIcon = (name: string) => {
-    const iconMap: { [name: string]: StaticImageData } = {
+  const renderTopicIcon = (name: InterestType) => {
+    const iconMap: Record<InterestType, StaticImageData> = {
       react: ReactIcon,
       rust: RustIcon,
       javascript: JavascriptIcon,
@@ -71,7 +53,7 @@ const LanguagePill = ({ topic, classNames, onClick }: LanguagePillProps) => {
       clojure: CloJureIcon,
     };
 
-    return iconMap[name] || "";
+    return iconMap[name];
   };
 
   return (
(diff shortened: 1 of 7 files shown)


lib/utils/getInterestOptions.ts
```diff
@@ -19,7 +19,7 @@ const interests = [
   "hacktoberfest",
   "clojure",
 ] as const;
-export type interestsType = (typeof interests)[number];
+export type InterestType = (typeof interests)[number];
```
nickytonline, line 22: Renamed this to Pascal case as it's a type. This avoids confusion with variables in the code.

components/atoms/LanguagePill/LanguagePill.tsx
```diff
@@ -21,35 +21,17 @@ import HacktoberfestIcon from "img/icons/interests/hacktoberfest.svg";
...
+  topic: InterestType;
   classNames?: string;
   onClick?: () => void;
 }
+
 const LanguagePill = ({ topic, classNames, onClick }: LanguagePillProps) => {
-  const renderTopicIcon = (name: string) => {
-    const iconMap: { [name: string]: StaticImageData } = {
+  const renderTopicIcon = (name: InterestType) => {
+    const iconMap: Record<InterestType, StaticImageData> = {
```
nickytonline, line 34: [Narrow the type](https://www.typescriptlang.org/docs/handbook/2/narrowing.html) using `InterestType` instead of `string`.

components/atoms/LanguagePill/LanguagePill.tsx
```diff
@@ -71,7 +53,7 @@ const LanguagePill = ({ topic, classNames, onClick }: LanguagePillProps) => {
       clojure: CloJureIcon,
     };
 
-    return iconMap[name] || "";
+    return iconMap[name];
```
nickytonline, line 56: Is the empty string necessary here @OgDev-01? If so, I'll put it back.

components/molecules/FilterHeader/filter-header.tsx
```diff
@@ -11,7 +11,7 @@ import useFilterPrefetch from "lib/hooks/useFilterPrefetch";
 import topicNameFormatting from "lib/utils/topic-name-formatting";
 import FilterCardSelect from "components/molecules/FilterCardSelect/filter-card-select";
 import getTopicThumbnail from "lib/utils/getTopicThumbnail";
-import { interestsType } from "lib/utils/getInterestOptions";
+import { InterestType } from "lib/utils/getInterestOptions";
```
nickytonline, line 14: There's a bunch of these which are just the name change for the type.

OgDev-01 (approved):

components/atoms/LanguagePill/LanguagePill.tsx
OgDev-01: not needed anymore with this refactor

PR #2215, opened 2023-11-27
fix: now list activity graphs only load data client-side

nickytonline:
## Description

Now only list info is loaded on the list activity page. All graph data is pulled on the client-side only now.

Here's a public list loading data for graphs client-side only now, https://deploy-preview-2215--oss-insights.netlify.app/lists/8b371419-6219-4448-9de1-f6293a25f98e/activity?range=30

## What type of PR is this? (check all applicable)

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Tickets & Documents

Fixes #2213

## Mobile & Desktop Screenshots/Recordings

https://github.com/open-sauced/app/assets/833231/75e7e0ae-3c7f-47f1-bff0-98b9743d06f8

## Added tests?

- [ ] 👍 yes
- [x] 🙅 no, because they aren't needed
- [ ] 🙋 no, because I need help

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook
- [x] 🙅 no documentation needed

PR #2241, opened 2023-12-01
fix: fixed crash on list contributor grid when in card view changing page size

### LINKED ISSUE #2240: Bug: Contributor table in a list crashes in card view when you change the page size to any value other than the current value
### Describe the bug

Contributor table in a list crashes when in card view when you change the page size to any value other than the current value.

### Steps to reproduce

1. For any list, go to the contributors tab.
2. The grid loads fine.
3. Switch to card view and the grid refreshes in card view.
4. Change the page limit to any other value than the currently selected page size will cause the page to crash on the client-side with Next.js's client-side error page.

image

### Browsers

_No response_

_No response_

### Code of Conduct

### Contributing Docs

nickytonline:
## Description

I've added react-error-boundary to the project that allows us to catch errors at the component level and display a fallback component if an error does occur.

Aside from that, it appears the initial contributor list data was causing issues with the component. This was working before, so not sure why removing it now fixes the issue, but as we get larger lists of contributors, it makes sense to offload this to the client-side only, much like the fix in #2215.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Fixes #2240

**Before**

The before view is running the dev server which is why you get the detailed error. If it was production, it would be the classic Next.js black screen of death.

https://github.com/open-sauced/app/assets/833231/8c65a80a-9853-4566-8b62-599d2bb8ec62

**After**

https://github.com/open-sauced/app/assets/833231/e66356c8-4b70-4818-8992-e038dd24c97d

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "react-day-picker": "^8.7.1",
     "react-dom": "^18.2.0",
     "react-emoji-render": "^2.0.1",
+    "react-error-boundary": "^4.0.11",
     "react-hot-toast": "^2.4.0",
     "react-icons": "^4.11.0",
     "react-loading-skeleton": "^3.1.0",
--- a/pages/lists/[listId]/contributors.tsx
+++ b/pages/lists/[listId]/contributors.tsx
@@ -2,6 +2,7 @@ import { GetServerSidePropsContext } from "next";
 import { createPagesServerClient } from "@supabase/auth-helpers-nextjs";
 import { useRouter } from "next/router";
 
+import { ErrorBoundary } from "react-error-boundary";
 import ListPageLayout from "layouts/lists";
 import { fetchApiData, validateListPath } from "helpers/fetchApiData";
 import Error from "components/atoms/Error/Error";
@@ -78,13 +79,17 @@ const ContributorsListPage = ({ list, initialData, isError, isOwner }: Contribut
       {isError ? (
         <Error errorMessage="Unable to load list of contributors" />
       ) : (
-        <ContributorsList
-          contributors={contributors}
-          meta={meta}
-          isLoading={isLoading}
-          setPage={setPage}
-          range={String(range) ?? "30"}
-        />
+        <ErrorBoundary
+          fallback={<div className="grid place-content-center">Error loading the list of contributors</div>}
+        >
+          <ContributorsList
+            contributors={contributors}
+            meta={meta}
+            isLoading={isLoading}
+            setPage={setPage}
+            range={String(range) ?? "30"}
+          />
+        </ErrorBoundary>
       )}
     </ListPageLayout>
   );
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -97,6 +97,7 @@
     "react-day-picker": "^8.7.1",
     "react-dom": "^18.2.0",
     "react-emoji-render": "^2.0.1",
+    "react-error-boundary": "^4.0.11",
```
nickytonline, line 100: react-error-boundary will allow us to have a fallback view if we wrap a component with it that crashes.

pages/lists/[listId]/contributors.tsx
```diff
@@ -17,48 +18,38 @@ export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
...
-  const limit = rawLimit ?? "10"; // Can pull this from the querystring in the future
-  const [{ data, error: contributorListError }, { data: list, error }] = await Promise.all([
-    fetchApiData<PagedData<DBListContributor>>({
-      path: `lists/${listId}/contributors?limit=${limit}&range=${range ?? "30"}`,
-      bearerToken,
-      pathValidator: validateListPath,
-    }),
-    fetchApiData<DBList>({ path: `lists/${listId}`, bearerToken, pathValidator: validateListPath }),
-  ]);
+  const { data: list, error } = await fetchApiData<DBList>({
```
Comment (nickytonline) [inline pages/lists/[listId]/contributors.tsx:23]: I've removed server-side rendering contributor list data and only kept critical stuff to load the page, e.g. list info.

Something related to the initial data for the contributor list is what caused the issue with the card view. Still trying to work out why, but regardless it makes sense to only load this client-side going forward as mentioned in the PR description.

pages/lists/[listId]/contributors.tsx
```diff
@@ -78,13 +68,17 @@ const ContributorsListPage = ({ list, initialData, isError, isOwner }: Contribut
...
       ) : (
-        <ContributorsList
-          contributors={contributors}
-          meta={meta}
-          isLoading={isLoading}
-          setPage={setPage}
-          range={String(range) ?? "30"}
-        />
+        <ErrorBoundary
+          fallback={<div className="grid place-content-center">Error loading the list of contributors</div>}
```
Comment (nickytonline) [inline pages/lists/[listId]/contributors.tsx:72]: I went with a barebones fallback error message @isabensusan, but going forward a bit of styling for this would be nice as we can use this for graphs or other data driven components.

OgDev-01 (approved):
LGTM codewise 👍

brandonroberts (approved):

PR #2265, opened 2023-12-06
fix: now contributors without language get accurate interest fallback

### LINKED ISSUE #2266: Bug: contributors table language is inacurate when interest fallback is used
### Describe the bug

Take a look at the image, and you'll see that some contributors have `J` as thier language.
image

This is incorrect, and after some digging, i found what the issue is.

`entries` here is `[][]` type return, and `contributors.interest.split(",")` would result into a single `[]`. so trying to merge both with a single `sort` and `map` return results to the issue when interests is used.
https://github.com/open-sauced/app/blob/fe00b876372e00c196d0a5c8c73c403a601cd9af/components/molecules/ContributorListTableRow/contributor-list-table-row.tsx#L51-L56

The conditional returns needs to be separated to solve this. 

### Steps to reproduce

1. Goto https://app.opensauced.pizza/1049/contributors/filter/vercel/next.js?range=30

### Browsers

Chrome

This was introduced in https://github.com/open-sauced/app/pull/2238

### Code of Conduct

### Contributing Docs

OgDev-01:
## Description
This PR fixes the bug with truncating contributors language when it used interest

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

fixes #2266 

### Before
image

### After
image

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/components/molecules/ContributorListTableRow/contributor-list-table-row.tsx
+++ b/components/molecules/ContributorListTableRow/contributor-list-table-row.tsx
@@ -52,8 +52,7 @@ function getTopContributorLanguages(contributor: DbUser) {
     return contributor.interests
       .split(",")
       .sort((a, b) => (a < b ? -1 : 1))
-      .slice(0, 2)
-      .map((language) => language.trim());
+      .slice(0, 2);
   }
 
   return entries
--- a/components/molecules/ContributorListTableRow/contributor-list-table-row.tsx
+++ b/components/molecules/ContributorListTableRow/contributor-list-table-row.tsx
@@ -52,6 +52,7 @@ function getTopContributorLanguages(contributor: DbUser) {
     return contributor.interests
       .split(",")
       .sort((a, b) => (a < b ? -1 : 1))
+      .slice(0, 2)
       .map((language) => language.trim());
   }
 
--- a/components/molecules/ContributorListTableRow/contributor-list-table-row.tsx
+++ b/components/molecules/ContributorListTableRow/contributor-list-table-row.tsx
@@ -11,6 +11,7 @@ import useContributorPullRequests from "lib/hooks/api/useContributorPullRequests
 import useRepoList from "lib/hooks/useRepoList";
 import { useFetchUser } from "lib/hooks/useFetchUser";
 import Checkbox from "components/atoms/Checkbox/checkbox";
+import { getTopContributorLanguages } from "lib/utils/contributor-utils";
 import { getActivity } from "../RepoRow/repo-row";
 import DevProfile from "../DevProfile/dev-profile";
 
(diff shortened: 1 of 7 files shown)


components/molecules/ContributorListTableRow/contributor-list-table-row.tsx
```diff
@@ -48,7 +48,14 @@ function getTopContributorLanguages(contributor: DbUser) {
   // some contributors will have empty language objects so we will pull their popular language from the interests field instead of defaulting to nothing
   const entries = Object.entries<string>(contributor.languages);
 
-  return (entries.length === 0 ? contributor.interests.split(",") : entries)
+  if (entries.length === 0) {
```
nickytonline, line 51: I’m on mobile at the moment, but is the only difference that we trim interests once they’re split now? I’m not sure how the trim changes much. Also, the new code returns all interests instead of two.

OgDev-01: the trim is just me being over cautious 😂. but the main change here is having interest sorted and sliced seperately from the entries since it ends up being an `[string, string][]` type

nickytonline: Not sure why that is because it should return one array or the other in the old code.

I'm still not sure why (condition ? array1 : array2) resulted in a tuple `[string, string]` though.

Anyways, mind testing to be certain all is good when there are no most used languages and interests are picked up instead?

brandonroberts (commented):
Agree with Nick's comments

OgDev-01:
> Anyways, mind testing to be certain all is good when there are no most used languages and interests are picked up instead?

I've tested and confirmed this is working after the fix @nickytonline, feel free to test run when you have the time to review fully 👍

components/molecules/ContributorListTableRow/contributor-list-table-row.tsx
nickytonline: I see what the issue is with the `[string, string]`. It's because of getting the object entries.

We can do this instead.

```typescript
function getTopContributorLanguages(contributor: DbUser) {
  // some contributors will have empty language objects so we will pull their popular language from the interests field instead of defaulting to nothing
  const entries = Object.entries<string>(contributor.languages);
  const languages = entries.length > 0 ? entries.map(([language]) => language) : contributor.interests.split(",");

  return languages.sort((a, b) => (a < b ? -1 : 1)).slice(0, 2);
}
```

I'm going to make that change and also add a unit test.

nickytonline (approved):
Giving it an approval, but if you can take a quick peek @jpmcb since I added code to this.

jpmcb (approved):
👏🏼

PR #2521, opened 2024-01-23
feat: V2 API top-project-contributions graph

jpmcb:
## Description

- Updates useState / onClick to use `repoName` for v2 API's top project contributions graph
- Update contributions-by-projects graph endpoint

## What type of PR is this? (check all applicable)

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Tickets & Documents

Closes: https://github.com/open-sauced/app/issues/2511
Closes: https://github.com/open-sauced/app/issues/2512

Related to: https://github.com/open-sauced/api/pull/488
Related to: https://github.com/open-sauced/api/pull/487

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook
- [ ] 🙅 no documentation needed

PR #2532, opened 2024-01-25
feat: fast forward `alpha` onto `beta`

jpmcb:


jpmcb:
Looks like we will need to add an additional commit to fixup the treemap changes that got squashed away

```
1:21:51 PM: Type error: Cannot find module "components/molecules/TreemapPrototype/special-node" or its corresponding type declarations.
```

Note: here's where we originally changed this on `alpha`: https://github.com/open-sauced/app/pull/2521

nickytonline (approved):
:shipit:

PR #2499, opened 2024-01-19
fix: Adding new repos to insights on edit page will trigger another API call

aryanas159:
## Description
fixes #2475
This PR fixes the issue, which was caused because when adding new repositories on the insights page, the response for each API call didn't return the number of PRs, so I used the useRepositories hook which reinitiates the API call every time the repos are added or removed. I did my best to solve this problem, but it might not be the best solution, I am willing to make any further changes required to fix this issue.

## What type of PR is this? (check all applicable)

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Mobile & Desktop Screenshots/Recordings

https://github.com/open-sauced/app/assets/114330931/b8dfae50-53a6-4e7a-ad6d-026e8ddc246a

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook
- [ ] 🙅 no documentation needed

PR #2550, opened 2024-01-29
Revert "fix: Adding new repos to insights on edit page will trigger another API call"

brandonroberts:
Reverts open-sauced/app#2499

Not sure why the build didn't continue to fail with the type error mentioned, but we need to revisit the usage in the edit page for an insight


jpmcb (approved):

nickytonline (approved):
🔄 :shipit:

PR #2552, opened 2024-01-30
feat: created the day range picker component

### LINKED ISSUE #2553: Feature: New day range picker
As a user of the workspaces dashboard, I want a day range picker that has three possible values: `7`, `30`, and `90` days so that I can filter repositories for a given day range. If a value isn't specified, it will default to `30`.

The component should have a label that says `Period:`

CleanShot 2024-01-29 at 20 02 20

CleanShot 2024-01-29 at 20 02 43

* [Designs in Figma](https://www.figma.com/file/sqRXS7hw6E93QgRMCl3WYz/Workspaces?type=design&node-id=902-16424&mode=dev)
* Relates to #2438

nickytonline:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Closes #2553

CleanShot 2024-01-29 at 19 48 45

## Steps to QA

N/A [currently it's only in Storybook](https://deploy-preview-2552--design-insights.netlify.app/?path=/docs/components-shared-day-range-picker--docs)

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [x] 📕 storybook

### DIFF
--- a/components/atoms/Select/select.tsx
+++ b/components/atoms/Select/select.tsx
@@ -19,7 +19,7 @@ const SelectTrigger = React.forwardRef<
   <SelectPrimitive.Trigger
     ref={ref}
     className={clsx(
-      "flex min-h-[2.1rem] w-full items-center radix-select-trigger justify-between rounded-lg overflow-hidden  px-3 py-1.5 text-sm border  focus:outline-none  disabled:cursor-not-allowed disabled:opacity-50",
+      "flex min-h-[2.1rem] items-center radix-select-trigger justify-between rounded-lg overflow-hidden  px-3 py-1.5 text-sm border  focus:outline-none  disabled:cursor-not-allowed disabled:opacity-50",
       className
     )}
     {...props}
--- a/components/atoms/Select/single-select.tsx
+++ b/components/atoms/Select/single-select.tsx
@@ -2,6 +2,7 @@ import React, { useEffect, useRef, useState } from "react";
 import { RiArrowDownSLine } from "react-icons/ri";
 import { BiSearch } from "react-icons/bi";
 
+import clsx from "clsx";
 import { truncateString } from "lib/utils/truncate-string";
 
 import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from "./select";
@@ -14,6 +15,7 @@ interface SingleSelectProps {
   options: { label: string; value: string }[];
   position?: "popper" | "item-aligned";
   isSearchable?: boolean;
+  insetLabel?: string;
 }
 
 const SingleSelect = ({
@@ -24,6 +26,7 @@ const SingleSelect = ({
   position,
   inputPlaceholder,
   isSearchable = false,
+  insetLabel,
 }: SingleSelectProps) => {
   const inputRef = useRef<HTMLInputElement>(null);
   const [inputValue, setInputValue] = useState("");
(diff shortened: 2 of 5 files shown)


components/shared/DayRangePicker.tsx
```diff
@@ -0,0 +1,45 @@
...
+  },
+];
+
+interface DayRangePickerProps {
+  onDayRangeChanged?: (value: string) => void;
+}
+
+export const DayRangePicker = ({ onDayRangeChanged }: DayRangePickerProps) => {
+  const router = useRouter();
+  const dayRange = (Number(router.query.range) || 30) as DayRange;
```
nickytonline, line 27: I'm curious on your thoughts @brandonroberts @jpmcb about this design decision. In other places we pass in the range or other query string param value to a filter component, but in this case I'm not doing that.

The reasoning being is the job of this component is to allow you to pick a day range and then it adds the day range to the querystring of the page it's currently on. This way, it's the pages responsibility to handle the updated query string value, `range` in this case to load data based on that value.

TLDR; it forces URL driven development. 😎

components/shared/DayRangePicker.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  const onDefaultDayRangeChanged = (value: string) => {
+    router.push({ query: { ...router.query, range: value } });
+    onDayRangeChanged?.(value);
```
nickytonline, line 31: I made this an optional prop as the component handles all the logic for routing/querystring, but someone might still want to do something. Maybe we don't even need this atm because maybe [YAGNI](https://codibly.com/news-insights/yagni-how-to-do-things-when-you-actually-need-them-to-be-done/).

components/shared/DayRangePicker.tsx
```diff
@@ -0,0 +1,45 @@
...
+  };
+
+  return (
+    <label className="w-max font-semibold">
```
nickytonline, line 35: A label is present for a11y reasons, but only viewable by a screen reader as the inset label is what a sighted user would see which matches with @isabensusan's design.

components/atoms/Select/select.tsx
```diff
@@ -19,7 +19,7 @@ const SelectTrigger = React.forwardRef<
   <SelectPrimitive.Trigger
     ref={ref}
     className={clsx(
-      "flex min-h-[2.1rem] w-full items-center radix-select-trigger justify-between rounded-lg overflow-hidden  px-3 py-1.5 text-sm border  focus:outline-none  disabled:cursor-not-allowed disabled:opacity-50",
+      "flex min-h-[2.1rem] items-center radix-select-trigger justify-between rounded-lg overflow-hidden  px-3 py-1.5 text-sm border  focus:outline-none  disabled:cursor-not-allowed disabled:opacity-50",
```
nickytonline, line 22: I navigated around the site where the `<Select />` component is used and removing `w-full` didn't change anything appearance wise. I removed it so that I could style the day range picker properly.

components/atoms/Select/single-select.tsx
```diff
@@ -51,7 +54,11 @@ const SingleSelect = ({
       }}
     >
       <SelectTrigger
-        className="bg-white min-w-[200px] w-full border focus-within:border-orange-500 focus-within:ring-orange-100 focus-within:ring ring-light-slate-6"
+        data-inset-label={insetLabel}
+        className={clsx(
+          `bg-white border`,
+          insetLabel && `before:content-[attr(data-inset-label)] before:mr-1 before:font-normal before:text-slate-500`
```
nickytonline, line 60: Only add the inset label to the combobox's [::before](https://developer.mozilla.org/en-US/docs/Web/CSS/::before) if the `insetLAbel` prop is set.

bdougie (approved):

components/shared/DayRangePicker.tsx
brandonroberts: I think its reasonable as it only controls a query param

brandonroberts (approved):

PR #2608, opened 2024-02-07
feat: add repositories to workspace in bulk

### LINKED ISSUE #2603: Feature: Ability to paste a list of repositories to track
As a user of workspaces, I want to be able to paste a list of repositories that can be tracked.

Note these are rough mocks

CleanShot 2024-02-06 at 16 24 24

CleanShot 2024-02-06 at 16 28 20

CleanShot 2024-02-06 at 16 26 29

zeucapua:
## Description

Add a new option to `TrackedReposWizard` to allow users to add repositories by pasting them into a text area, each separated into new lines. User can then filter from the added repositories before finalizing the import.

NOTE: repositories are case sensitive when trying to fetch during `Update Workspace`. I added a note on the modal above where the text are so users can see before pasting their list. This means that if there is even only 'one' repo with the wrong cAse, it'll error out adding the entire list. How should we address this? @open-sauced/engineering 

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Closes #2603 

https://github.com/open-sauced/app/assets/20603494/0bd8f8f9-6056-48cf-81dd-a0faba11d2b1

## Steps to QA

1. Go to `/workspaces/new` or `/workspaces/settings`
2. Click `Add repositories`
3. Paste a list of repositories (CASE SENSITIVE)
4. Filter through the pasted repositories
5. Finish modal and click `Update Workspace`

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

list-gif

### DIFF
--- a/components/Workspaces/TrackedRepoWizard/FilterPastedReposStep.tsx
+++ b/components/Workspaces/TrackedRepoWizard/FilterPastedReposStep.tsx
@@ -1,4 +1,3 @@
-import { FaSearch } from "react-icons/fa";
 import { useEffect, useState } from "react";
 import { SearchedReposTable } from "../SearchReposTable";
 
@@ -7,20 +6,6 @@ interface FilterPastedReposStepProps {
   onToggleAllRepos: (checked: boolean) => void;
   repositories: Map<string, boolean>;
 }
-
-const EmptyState = () => {
-  return (
-    <div className="grid place-content-center">
-      <div className="text-center flex flex-col items-center p-24 gap-2 max-w-lg">
-        <div className="p-3 border rounded-lg mb-2">
-          <FaSearch size={24} className="text-light-slate-9" />
-        </div>
-        <span className="font-semibold">No repositories added yet!</span>
-        <span>Use the search bar to find the repositories you want to track on your workspace.</span>
-      </div>
-    </div>
-  );
-};
 
 export const FilterPastedReposStep = ({ onToggleRepo, onToggleAllRepos, repositories }: FilterPastedReposStepProps) => {
   const [filteredRepositories, setFilteredRepositories] = useState<Map<string, boolean>>(new Map());
@@ -43,17 +28,13 @@ export const FilterPastedReposStep = ({ onToggleRepo, onToggleAllRepos, reposito
 
   return (
     <div className="flex flex-col gap-4 h-96 max-h-96">
-      {repositories.size === 0 ? (
-        <EmptyState />
-      ) : (
-        <SearchedReposTable
-          type="by-repos"
-          repositories={filteredRepositories}
-          onFilter={onFilterRepos}
-          onToggleRepo={onToggleRepo}
-          onToggleAllRepos={onToggleAllRepos}
-        />
-      )}
+      <SearchedReposTable
+        type="by-repos"
+        repositories={filteredRepositories}
+        onFilter={onFilterRepos}
+        onToggleRepo={onToggleRepo}
+        onToggleAllRepos={onToggleAllRepos}
+      />
     </div>
   );
 };
--- a/components/Workspaces/TrackedRepoWizard/FilterPastedReposStep.tsx
+++ b/components/Workspaces/TrackedRepoWizard/FilterPastedReposStep.tsx
@@ -1,0 +1,59 @@
+import { FaSearch } from "react-icons/fa";
+import { useEffect, useState } from "react";
+import { SearchedReposTable } from "../SearchReposTable";
+
+interface FilterPastedReposStepProps {
+  onToggleRepo: (repo: string, isSelected: boolean) => void;
+  onToggleAllRepos: (checked: boolean) => void;
+  repositories: Map<string, boolean>;
+}
+
+const EmptyState = () => {
+  return (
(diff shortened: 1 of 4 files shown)


nickytonline:
Let us know if this icon is OK @isabensusan 

CleanShot 2024-02-06 at 19 50 53

bdougie:
Reading the note about the cAse.

I am curious if there is a need for case sensitivity and if we can lowercase (sanitize) it all to remove the burden from the user.

jpmcb:
From the APIs perspective, we do sanitize case sensitivity, so when adding repos, it shouldn't be necessary to have the users make repos case sensitive. And from GitHub's perspective, repo names (and most URL parameters, like usernames) are also not case sensitive. Example: https://github.com/Open-Sauced/APP is the same as https://github.com/open-sauced/app

jpmcb:
If there's an error when attempting to fetch repos during update workspace with case sensitive repos, that sounds like a bug 🐛

zeucapua:
Went to the preview deploy, now I'm having different behavior. It's still not adding not correctly cased repositories, but it's adding those that are. 

See video, only `TanStack/router` is correctly cased. Clearly the table gets all the repos, just on `Update Workspace` something is going wrong. I can look into it in the morning.

https://github.com/open-sauced/app/assets/20603494/343300a4-2740-4447-913f-f9fb4b5c7e47

components/Workspaces/TrackedRepoWizard/PasteReposStep.tsx
```diff
@@ -0,0 +1,31 @@
...
+import Button from "components/atoms/Button/button";
+
+interface PasteReposStepProps {
+  onBulkAddRepos: (repos: string[]) => void;
+}
+
+export const PasteReposStep = ({ onBulkAddRepos }: PasteReposStepProps) => {
+  const [pastedInput, setPastedInput] = useState("");
+
+  const parseInput = () => {
```
nickytonline, line 11: refactor (blocking): if the textarea is empty return instead of doing the bulk import. Otherwise you get to the next step with a repository that is an empty string.

CleanShot 2024-02-07 at 09 51 22

Until the value of the textarea has text, the import button should be disabled.

components/Workspaces/TrackedRepoWizard/PasteReposStep.tsx
```diff
@@ -0,0 +1,31 @@
...
+  };
+
+  return (
+    <div className="flex flex-col gap-4 h-96 max-h-96">
+      <p className="text-red-500">NOTE: repository names are case sensitive</p>
+      <textarea
+        value={pastedInput}
+        onChange={(e) => setPastedInput(e.target.value)}
+        placeholder="Paste repositories here line by line (e.g. open-sauced/app)"
+        className="p-4 border rounded-xl h-full"
```
nickytonline, line 24: When the textarea gets focus, it uses the default blue outline of the browser. There are custom focus styles already set in the application. It's a little more involved to add it to the global styles for now, so in the interim, you can add the focus styles to the textarea in this component.

```
outline-none focus-visible:ring focus-visible:border-orange-500 focus-visible:ring-orange-100
```

nickytonline (changes requested):
This is working great @zeucapua! A couple change requests.

nickytonline:
I've ignored the casing issue mentioned as it looks like @jpmcb is seeing if it's an issue in the API.

nickytonline:
> I opened a PR in the API with a fix for normalizing the case when searching for repos.
> 
> 
> 
> https://github.com/open-sauced/api/pull/547

Cool cool. So we can remove the red text warning then @zeucapua

zeucapua:
Added validation and sanitation checks to:
- disable importing if the textarea is empty (even with empty lines)
- remove any empty lines between the repos that can result in an empty fetch

Also changed styling so the case sensitivity note is gone and the textarea on focus is more on brand. Attached a video to see it all in action:

https://github.com/open-sauced/app/assets/20603494/fb5a709f-ee2c-4317-a54e-878ff0633aea

components/Workspaces/TrackedRepoWizard/FilterPastedReposStep.tsx
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    setFilteredRepositories(updates);
+  };
+
+  return (
+    <div className="flex flex-col gap-4 h-96 max-h-96">
+      {repositories.size === 0 ? (
+        <EmptyState />
+      ) : (
```
nickytonline, line 48: I don't think the empty state us required as we'll always be coming from the pasted textarea that has at least one repo in it.

nickytonline (approved):
Works great @zeucapua! thanks for making the changes!

🚀

brandonroberts (approved):
🍕

PR #3094, opened 2024-04-02
fix: removed unused parts of metric card for now

### LINKED ISSUE #3095: Bug: Removed unused portions of metric card for now.
We don't need the ellipse button menu at the moment or the percentage in the bottom left corner.

CleanShot 2024-04-02 at 16 56 29

nickytonline:

Fixes #3095

CleanShot 2024-04-03 at 10 39 49

**Before**

CleanShot 2024-04-03 at 10 27 19

**After**

CleanShot 2024-04-03 at 10 26 35

## Steps to QA

1. Visit `/s/analogjs/analog`
2. Notice the metric cards have the ellipse button menu and percentage in the bottom left corner removed.

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/components/Graphs/MetricCard.stories.tsx
+++ b/components/Graphs/MetricCard.stories.tsx
@@ -1,5 +1,4 @@
 import { Meta, StoryObj } from "@storybook/react";
-import { useFetchMetricStats } from "lib/hooks/api/useFetchMetricStats";
 import MetricCard from "./MetricCard";
 
 type MetricCardAndRepositoryArgs = React.ComponentProps<typeof MetricCard> & { repository: string };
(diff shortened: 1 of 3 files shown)


nickytonline (commented):
@isabensusan, the height of these cards is a little less now. Would you like to keep the same heigh even if the percentage is no longer there?

brandonroberts (approved):

zeucapua:
While we're making changes to the Metric Cards, we should rename these to just "Stars"/"Forks" since multiple people have expressed confusion on the name when the number reflects the total for the range period.

nickytonline:
> While we're making changes to the Metric Cards, we should rename these to just "Stars"/"Forks" since multiple people have expressed confusion on the name when the number reflects the total for the range period.

Maybe `Stars/Forks over x days`?

Thoughts @bdougie?

bdougie:
> Maybe Stars/Forks over x days?

This makes more sense.

isabensusan:
could we add the time period selected next to the title on the card? like so:

`font-medium text-base text-slate-500`

image

and for the percentages, I know that it's not functional right now, but could we make it work? the idea is to give the user a comparison vs the last period. So if I'm seeing stars for the past 30 days, the percentage should be calculated on the difference with the previous 30 days.

We could have it on a follow up PR too

nickytonline:
> could we add the time period selected next to the title on the card? like so:
> 
> `font-medium text-base text-slate-500`
> 
> image
> and for the percentages, I know that it's not functional right now, but could we make it work? the idea is to give the user a comparison vs the last period. So if I'm seeing stars for the past 30 days, the percentage should be calculated on the difference with the previous 30 days.
> 
> We could have it on a follow up PR too

Done @isabensusan. ✅

PR #3108, opened 2024-04-03
feat: added million lint to the project

nickytonline:
## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [ ] Tier 4

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,3 +1,4 @@
+const MillionLint = require("@million/lint");
 const interests = [
   "javascript",
   "python",
@@ -23,7 +24,7 @@ const interests = [
 ];
 
 /** @type {import('next').NextConfig} */
-module.exports = {
+module.exports = MillionLint.next()({
   productionBrowserSourceMaps: true,
   reactStrictMode: true,
   images: {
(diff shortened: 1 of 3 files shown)


next.config.js
```diff
@@ -23,7 +24,7 @@ const interests = [
 ];
 
 /** @type {import('next').NextConfig} */
-module.exports = {
+module.exports = MillionLint.next()({
```
nickytonline, line 27: `MillionLint.next()` creates a function that wraps the Next.js config to bootstrap Million Lint. When it builds for production, it will not bootstrap Million Lint.

bdougie:
I missed the part about needing the VScode extension and saw this. Should we add some context to the README for future contributors? Is it also possible 

Screen Shot 2024-04-03 at 4 51 50 PM

The error message is helpful, but I imagine we will have a few folks ping the discord. Also realize they literally just had their YC demo day and probably going to scale fast as well, but defer to the rest of the engineers to approve and provide feedback.

nickytonline:
> I missed the part about needing the VScode extension and saw this. Should we add some context to the README for future contributors? 
> 
> 
> 
> Screen Shot 2024-04-03 at 4 51 50 PM
> 
> 
> 
> The error message is helpful, but I imagine we will have a few folks ping the discord. Also realize they literally just had their YC demo day and probably going to scale fast as well, but defer to the rest of the engineers to approve and provide feedback. 
> 
> 

Yeah, good call about the readme. Also, happy to let this sit for a bit. Currently, I'm using it locally and it's been super helpful already.

bdougie:
> Also, there is no rush on this as it's not a direct high priority although it does relate to performance improvements.

FWIW, I was happy to try and install it. This also directly correlates with our Q2 OKRs as well.

jpmcb:
+1 - this is a soft blocker for me since I don't use neovim (but I suppose I can always spin up vscode to run the app). But, would prefer our code artifacts to have as minimal vendor lock in as possible

nickytonline:
> +1 - this is a soft blocker for me since I don't use neovim (but I suppose I can always spin up vscode to run the app). But, would prefer our code artifacts to have as minimal vendor lock in as possible

We can use an environment variable if people want to opt in and I can load it based on that, but I opened an issue for a silent mode. See https://github.com/aidenybai/million/issues/1015

nickytonline:
> > +1 - this is a soft blocker for me since I don't use neovim (but I suppose I can always spin up vscode to run the app). But, would prefer our code artifacts to have as minimal vendor lock in as possible
> 
> We can use an environment variable if people want to opt in and I can load it based on that, but I opened an issue for a silent mode. See [aidenybai/million#1015](https://github.com/aidenybai/million/issues/1015)

This is now using an environment variable @jpmcb.

README.md
```diff
@@ -47,6 +47,12 @@ To start a local copy of the app on port `3000`:
 npm run dev
 ```
 
+Interested in helping with potential performance problems? Run the dev server with [Million Lint](https://million.dev/lint). Note that Million Lint is currently only supported for VS Code and it requires the [Million Lint extension](https://marketplace.visualstudio.com/items?itemName=million.million-lint).
```
nickytonline, line 50: Does this wording sound OK @BekahHW?

next.config.js
```diff
@@ -1,3 +1,6 @@
+const MillionLint = require("@million/lint");
+const millionLint = process.env.USE_MILLION_LINT ? MillionLint.next() : (config) => config;
```
brandonroberts, line 2: Does it matter if we only import Million Lint when the environment variable is set?

```suggestion
const millionLint = process.env.USE_MILLION_LINT ? require("@million/lint").next() : (config) => config;
```

nickytonline: That should work too. I can make the adjustment. I imagine they have a guard for not running in prod, but no point in loading it if you're not using it.

brandonroberts (approved):

jpmcb (approved):
🧹 🧹 🧹

PR #3139, opened 2024-04-05
feat: Integrate Grafana Faro

jpmcb:
## Description

Adds the react integrations and Faro error boundary around the app.
Initializes the default tracing/web integrations for Faro to get us some internal metrics and tracing.

N/a - related to ongoing observability and performance work

Here's a snapshot after running this locally and getting some of those metrics into our Grafana:

Screenshot 2024-04-05 at 9 53 57 AM

## Steps to QA

1. Run app locally
2. Modify the environment variables to use the collector and set the env to "beta" to bypass check

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

Need to update Netlify env vars to use these

### DIFF
--- a/lib/utils/grafana.ts
+++ b/lib/utils/grafana.ts
@@ -16,13 +16,15 @@ export const initGrafanaFaro = () => {
 
     instrumentations: [
       // load the mandatory web instrumentation
-      ...getWebInstrumentations(),
+      ...getWebInstrumentations({
+        captureConsole: true,
+      }),
 
       // add tracing instrumentation which should include the React Profiler
       new TracingInstrumentation(),
 
       new ReactIntegration({
-        // In the future, we may choose to integrate with the router instrumentation to
+        // In the future, we may choose to integrate with React router instrumentation to
         // get deeper metrics on matched routes, navigation types, etc.
         // Next/router doesn't seem to be supported which won't give us route metrics.
         //
--- a/lib/utils/grafana.ts
+++ b/lib/utils/grafana.ts
@@ -1,0 +1,31 @@
+import { getWebInstrumentations, initializeFaro, ReactIntegration } from "@grafana/faro-react";
+import { TracingInstrumentation } from "@grafana/faro-web-tracing";
+
+export const initGrafanaFaro = () => {
+  const faro = initializeFaro({
+    url: process.env.NEXT_PUBLIC_FARO_COLLECTOR_URL,
+
+    app: {
+      name: process.env.NEXT_PUBLIC_FARO_APP_NAME,
+      version: "1.0.0",
+      environment: process.env.NEXT_PUBLIC_FARO_ENVIRONMENT,
+    },
+
+    instrumentations: [
+      // load the mandatory web instrumentation
+      ...getWebInstrumentations(),
+
+      // add tracing instrumentation which should include the React Profiler
+      new TracingInstrumentation(),
+
+      new ReactIntegration({
+        // In the future, we may choose to integrate with the router integration to
+        // get deeper metrics on matched routes, navigation types, etc.
+        // Reference: https://github.com/grafana/faro-web-sdk/tree/main/packages/react
+        // router: {}
+      }),
+    ],
+  });
+
+  return faro;
+};
--- a/lib/utils/grafana.ts
+++ b/lib/utils/grafana.ts
@@ -1,13 +1,16 @@
 import { getWebInstrumentations, initializeFaro, ReactIntegration } from "@grafana/faro-react";
 import { TracingInstrumentation } from "@grafana/faro-web-tracing";
 
+// Get the current, runtime version of the App to surface to Faro
+import packageJson from "../../package.json";
+
 export const initGrafanaFaro = () => {
-  const faro = initializeFaro({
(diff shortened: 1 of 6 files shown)


pages/_app.tsx
```diff
@@ -62,6 +64,12 @@ type ComponentWithPageLayout = AppProps & {
 };
 
 function MyApp({ Component, pageProps }: ComponentWithPageLayout) {
+  useEffect(() => {
+    if (process.env.NEXT_PUBLIC_FARO_COLLECTOR_URL && process.env.NEXT_PUBLIC_FARO_APP_ENVIRONMENT != "local") {
+      initGrafanaFaro();
+    }
+  }, []);
+
```
jpmcb, line 72: I'm unsure this is doing what I expect it to do: Faro needs to initialize on the client (not the server) so I figured that a `userEffect` would get the app to 

But it only seems to capture traces/profiles when the App first loads on the `/feed` and I see this log go through:

```
Faro
 Faro is already registered. Either add instrumentations, transports etc. to the global faro instance or use the "isolate" property
```

which makes me think that maybe there's a different way to instantiate it - Investigating.

jpmcb: Ended up going with a `useRef` which now doesn't seem to cause Faro to be re-initilized

pages/_app.tsx
```diff
@@ -151,13 +165,15 @@ function MyApp({ Component, pageProps }: ComponentWithPageLayout) {
         <SessionContextProvider supabaseClient={supabaseClient} initialSession={pageProps.initialSession}>
           <PostHogProvider client={posthog}>
             <TipProvider>
-              {Component.PageLayout ? (
-                <Component.PageLayout>
+              <FaroErrorBoundary>
```
jpmcb, line 168: I'm not sure this is the right place for the error boundary ... their React docs make it seem like this should wrap a rendered `<App />` and I'm not seeing errors flow through when testing.

lib/utils/grafana.ts
```diff
@@ -0,0 +1,35 @@
...
+      new TracingInstrumentation(),
+
+      new ReactIntegration({
+        // In the future, we may choose to integrate with the router instrumentation to
+        // get deeper metrics on matched routes, navigation types, etc.
+        // Next/router doesn't seem to be supported which won't give us route metrics.
+        //
+        // Reference: https://github.com/grafana/faro-web-sdk/tree/main/packages/react
+        //
+        // router: {}
```
jpmcb, line 31: I'm not getting any additional routes besides `feed/` in the traces and I _think_ it's because it's due to there note being a router config here (and not really seeing a way for it to support `next/router`, only React router)

jpmcb:
After experimenting abit with this, we can definitely get this in and start getting some good latency data. But there are some missing integrations in the library with NextJS that i think make it abit half baked. We can draft this for now and I can watch to see when they land some Next integrations.

pages/_app.tsx
nickytonline: What do the Faro docs recommend for a React app?

jpmcb:
Pushing logs and errors through is working 👍🏼 



I say let's get this in and we can keep an eye on their NextJS instrumentation coming through

pages/_app.tsx
```diff
@@ -62,6 +64,18 @@ type ComponentWithPageLayout = AppProps & {
...
+
+  useEffect(() => {
+    if (
+      process.env.NEXT_PUBLIC_FARO_COLLECTOR_URL &&
+      process.env.NEXT_PUBLIC_FARO_APP_ENVIRONMENT != "local" &&
+      !faroRef.current
+    ) {
+      faroRef.current = initGrafanaFaro();
+    }
+  }, []);
```
nickytonline, line 77: The useRef seems unnecessary. If all you're trying to do is initialize this for the client-side only, this should suffice.

```suggestion
  useEffect(() => {
    if (
      process.env.NEXT_PUBLIC_FARO_COLLECTOR_URL &&
      process.env.NEXT_PUBLIC_FARO_APP_ENVIRONMENT != "local"
    ) {
      initGrafanaFaro();
    }
  }, []);
```

jpmcb: Yeah, that's what I tried initially but would see this pop up in:

```
Faro
 Faro is already registered. Either add instrumentations, transports etc. to the global faro instance or use the "isolate" property
```

But that must be due to a re-render that happens? Either way, seems to be fine since the points still get sent to the collector endpoint ok:

Screenshot 2024-04-05 at 3 52 47 PM

erkattak:
I'm curious - is the entirety of your app client-side only, or are you doing server-side rendering of components where you're using faro?

jpmcb:
> is the entirety of your app client-side only, or are you doing server-side rendering of components where you're using faro?

We do have server side rendering. Unfortunately, Faro is not well supported for Next JS yet so we're uncertain the future of this integration.

jpmcb:
Closing as this has grown stale and there is still no adoption in Grafana Faro of NextJS. Let's continue to watch https://github.com/grafana/faro-web-sdk/issues/496 and see if there's any movement there.

PR #3176, opened 2024-04-12
feat: implement suggestion boxes for `StarSearch`

zeucapua:
## Description

Implement functionality for suggestion boxes. Users can click one and it will automatically prompt the AI. Also added a `Back to Start` button that returns the user from the chat history state to the initial state with the suggestion boxes.

## Tier (staff will fill in)

- [x] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [ ] Tier 4

greys-anatomy-amelia-shepherd-gif

### DIFF
--- a/pages/starsearch.tsx
+++ b/pages/starsearch.tsx
@@ -6,8 +6,8 @@ import Image from "next/image";
 import { getAllFeatureFlags } from "lib/utils/server/feature-flags";
 import Card from "components/atoms/Card/card";
 import ProfileLayout from "layouts/profile";
-import { ScrollArea } from "components/atoms/ScrollArea/scroll-area";
 import { getAvatarById } from "lib/utils/github";
+import Button from "components/shared/Button/button";
 
 export async function getServerSideProps(context: GetServerSidePropsContext) {
   const supabase = createPagesServerClient(context);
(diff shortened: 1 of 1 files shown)


brandonroberts:
Just a suggestion that when clicking on the suggestion box it fills in the question prompt instead of going directly to the search results. That way they can tweak the prompt before submitting

zeucapua:
> Just a suggestion that when clicking on the suggestion box it fills in the question prompt instead of going directly to the search results. That way they can tweak the prompt before submitting

We can do that! This is different from the initial plan where it would automatically prompt the AI. Which should we do? @isabensusan

pages/starsearch.tsx
```diff
@@ -115,9 +115,21 @@ export default function StarSearchPage({ userId, bearerToken }: StarSearchPagePr
   const renderState = () => {
     switch (starSearchState) {
       case "initial":
-        return <Header />;
+        return (
+          <section className="flex flex-col text-center items-center gap-4 lg:pt-24 z-10">
+            <Header />
+            <SuggestionBoxes onSubmitPrompt={submitPrompt} />
+          </section>
```
nickytonline, line 122: These don't need to be sections (here and below), a div is fine as it's presentational. The generic [section](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section) element, if used should almost always have a header. In this particular case, the element is being used as a container for styling.

```suggestion
          <div className="flex flex-col text-center items-center gap-4 lg:pt-24 z-10">
            <Header />
            <SuggestionBoxes onSubmitPrompt={submitPrompt} />
          </div>
```

nickytonline (changes requested):

FatumaA:
This is so cool 🤩

From the screenrecording, isn't it typical in chat interfaces to have one of the chatters chat box come from the opposite side? And have the 2 chatters chat box be slightly differently styled?

brandonroberts (approved):
Works as expected!

isabensusan (changes requested):
☝️ see comments

zeucapua:
It's not overlayed anymore, instead moving the chat above it. I can fix it more in the future, but figured this would be good to merge.
https://github.com/open-sauced/app/assets/20603494/7c79932d-95ee-4ea7-b251-9d6eb032cce0

isabensusan (approved):
I'm still getting some weird behaviours with things overlapping , but I think overall this is good to merge -- we can do follow up PRs for styling

image

image

PR #3194, opened 2024-04-16
fix: quick fix StarSearch response

zeucapua:
## Description

Better parsing for StarSearch response.

BEFORE
image

AFTER
image

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -27313,14 +27313,11 @@
       }
     },
     "node_modules/swr": {
-      "version": "2.0.3",
-      "resolved": "https://registry.npmjs.org/swr/-/swr-2.0.3.tgz",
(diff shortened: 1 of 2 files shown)


nickytonline (approved):
Giving the LGTM because we just paired on this.

🚢

bdougie:
We talked about this but it would be ideal to format order lists and add hyperlinks in a follow up or tomorrow

isabensusan:
This is definitely closer to what we want to achieve, no more white spaces!

On a follow up PR, it'd be great to add even more formatting so that bolded words actually render as bold, and we add some line breaks and lists format. The prompt reply seems to be coming in in markdown format (see screenshot), so rendering that would be great.

image

jpmcb (approved):
👍🏼

bdougie (approved):
🚀🚀🚀

PR #3305, opened 2024-05-02
fix: filter out undefined values for building workspace owners array

brandonroberts:
## Description

- Fixes a bug where a workspaces has owners and viewers, which causes an error in `getServerSideProps` when `undefined` is returned in the owners array

## Steps to QA

1. Navigate to a workspace with an owner and viewer
2. Go to a contributor insight
3. Note the page loads correctly

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/pages/workspaces/[workspaceId]/contributor-insights/[listId]/activity.tsx
+++ b/pages/workspaces/[workspaceId]/contributor-insights/[listId]/activity.tsx
@@ -84,7 +84,7 @@ export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
     if (member.role === "owner") {
       return member.member.login;
     }
-  });
+  }).filter(Boolean);
 
   const isOwner = !!(workspaceMembers?.data || []).find(
     (member) => member.role === "owner" && member.user_id === userId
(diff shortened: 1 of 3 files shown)


jpmcb (commented):
> when `undefined` is returned in the owners array

That's surprising since my assumption was that the `"members": []` would always have at _least_ one member/owner. I thought we had blocked removing all members from a workspace (or at least owners could not remove themselves)

Or maybe it's something else? I'm very curious why this would be getting undefined in the members array.

brandonroberts:
> > when `undefined` is returned in the owners array
> 
> That's surprising since my assumption was that the `"members": []` would always have at _least_ one member/owner. I thought we had blocked removing all members from a workspace (or at least owners could not remove themselves)
> 
> Or maybe it's something else? I'm very curious why this would be getting undefined in the members array.

The data is correct

```json
[
  { login: 'me', role: 'owner' },
  { login: 'you', role: 'viewer' },
]
```


```ts
  const owners = Array.from(workspaceMembers?.data || [], (member: { role: string; member: Record<string, any> }) => {
    if (member.role === "owner") {
      return member.member.login;
    }
  }).filter(Boolean);
```

If its an owner, the login is returned. If its not, undefined is returned. Then when `getServersideProps` tries to serialize the data

nickytonline (approved):
🚢

zeucapua (approved):

jpmcb (approved):
Thanks for explaining  !!!!!!!!!!  🧠 ◀️ 🥇

PR #2615, opened 2024-02-07
fix: Now you can't type more than 255 characters in the bio field of user settings

Pagla-Dasu:
Now, it Can't keep typing in the bio field of user settings after 255 characters. Changed the code in user-sttings-page.tsx

fix #2612

## Description

This PR fixes the issue #2612 fixing the maximum character issue of 255 characters.

## What type of PR is this? (check all applicable)

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Tickets & Documents

Please use this format link issue numbers: Fixes #2612 
https://github.com/open-sauced/app/issues/2612 

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook
- [x] 🙅 no documentation needed

Issue #2612, opened 2024-02-07
Bug: Can keep typing in the bio field of user settings after 255 characters


nickytonline:
We currently display a message saying the bio is too long in the user settings, but we're not leveraging the native [maxLength](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength) HTML attribute to prevent someone from typing past the limit.

![CleanShot 2024-02-07 at 08 15 24](https://github.com/open-sauced/app/assets/833231/a235aecb-170b-4efc-ac58-2a942677c3a0)

As well, since `maxLength` prevents you from typing or pasting text past 255 characters, the `bio too long` message will actually never appear so it can be removed from the component.

This is what happens when `maxLength="255"` is set.

![CleanShot 2024-02-07 at 08 08 43](https://github.com/open-sauced/app/assets/833231/5e22c1cb-51f5-46d2-baf9-77f4fbff3527)

nickytonline:
The file to change is https://github.com/open-sauced/app/blob/3d157ab69ae34353ec169f6c6cbd12744c199a98/components/organisms/UserSettingsPage/user-settings-page.tsx#L273-L298

Here's what needs to be done to fix this

```diff
            {/* Bio section */}
            <div className="flex flex-col gap-2">
              <label className="flex flex-col w-full text-sm">
                Bio
                <div className="flex-1 px-2 shadow-input border transition rounded-lg py-1 flex items-center bg-light-slate-4 disabled:cursor-not-allowed focus-within:border-light-orange-9">
                  <textarea
                    rows={4}
                    placeholder="Tell us about yourself."
                    name="bio"
                    className="w-full focus:outline-none placeholder:font-normal placeholder-slate-400 bg-inherit"
                    value={bio}
+					maxLength={255}
                    onChange={(e) => setBio(e.target.value)}
                  ></textarea>
                </div>
              </label>

-              {bio?.length > 255 ? (
-                <p aria-live="assertive" className="text-light-red-10 text-xs">
-                  Bio too long
-                </p>
-              ) : (
-                <p aria-live="polite" className="text-xs">
-                  {bio?.length}/255
-                </p>
-              )}
+              <p aria-live="polite" className="text-xs">
+                {bio?.length}/255
+              </p>
            </div>
```

lexlucid:
.take

Pagla-Dasu:
.take

HodaSalim:
.take

devharsh2k4:
[...]

PR #3321, opened 2024-05-06
fix : Can keep typing in the bio field of user settings after 255 characters 

devharsh2k4:
## Description
We currently display a message saying the bio is too long in the user settings, but we're not leveraging the native [maxLength](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength) HTML attribute to prevent someone from typing past the limit.

While the user will add 255 words it will show a text which says Bio is full and now it cant contain more words

This PR  removes a bug 

Please use this format link issue numbers: Fixes #2615 
https://github.com/open-sauced/app/issues/2612

Screenshot from 2024-05-07 00-47-39

## Steps to QA

1.Go to User settings 
2. Click in Bio
3. fill your bio

😊

### DIFF
--- a/components/organisms/UserSettingsPage/user-settings-page.tsx
+++ b/components/organisms/UserSettingsPage/user-settings-page.tsx
@@ -289,9 +289,15 @@ const UserSettingsPage = ({ user }: UserSettingsPageProps) => {
                 </div>
               </label>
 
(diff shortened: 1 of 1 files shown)


nickytonline:
Closing this @devharsh2k4 as it's already assigned to @Pagla-Dasu and they have a PR up for this.

PR #3326, opened 2024-05-07
fix: Star-search md rendering renders spacing from LLM

jpmcb:
## Description

This adds 2 things primarily to the star-search markdown rendering:

1. Empty `data:` frames are counted as newlines from the model. This is done by making the `\s` spacing capture in the regex optional _and_ not forcing a capture in the `results` capture-group via `.+`: instead, we allow for any captures (including zero) through `.*`.
2. When we get an empty string in the capture group, adds a "&nbsp; \n" which _seems_ to be the way to get react-markdown to render some newlines/spacing correctly. I may be wrong on this, but it seems to work _okay_.

This also changes `Markdown` --> `ReactMarkdown`. This seems to be the way the library recommends rendering "safe" markdown that can't have code injected into it: https://www.npmjs.com/package/react-markdown/v/8.0.6

I also noticed what seems to be some improvements to the rendering of the markdown switching to `ReactMarkdown`.

Before:

Screenshot 2024-05-06 at 10 44 16 PM

After:

Screenshot 2024-05-06 at 10 47 35 PM

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [x] Tier 3
- [ ] Tier 4

### DIFF
--- a/pages/star-search/index.tsx
+++ b/pages/star-search/index.tsx
@@ -4,7 +4,7 @@ import { MdOutlineSubdirectoryArrowRight } from "react-icons/md";
 import { useEffect, useRef, useState } from "react";
 
 import Image from "next/image";
-import Markdown from "react-markdown";
+import ReactMarkdown from "react-markdown";
 import { TrashIcon } from "@heroicons/react/24/outline";
 import { BsArrowUpShort } from "react-icons/bs";
 import { ThumbsdownIcon, ThumbsupIcon, XCircleIcon } from "@primer/octicons-react";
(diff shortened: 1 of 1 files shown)


jpmcb:
I also came across this plugin: https://github.com/remarkjs/remark-breaks which would add `<br>` to the content and maybe make it make more sense: it seems this is maybe what GitHub does in its markdown rendering for comments / issues, etc.

pages/star-search/index.tsx
```diff
@@ -4,7 +4,7 @@ import { MdOutlineSubdirectoryArrowRight } from "react-icons/md";
 import { useEffect, useRef, useState } from "react";
 
 import Image from "next/image";
-import Markdown from "react-markdown";
+import ReactMarkdown from "react-markdown";
```
nickytonline: The name doesn't matter as it's a default export. I put it back to what it was with your other changes and it still behaves the same/better with the new changes to formatting that you made elsewhere.

nickytonline (approved):
I pushed one change, but this is good to go.

:shipit:

pages/star-search/index.tsx
jpmcb: Makes sense, makes sense - thanks!!

PR #3343, opened 2024-05-08
feat: Lottery Factor chart for Repo pages

### LINKED ISSUE #3341: Feature: Lottery Factor Chart
## Overview
We want to build a lottery factor chart for our repo pages, so that users can see who the most active contributors are (based on commits) and the % of their contributions over the total.

## Endpoint
We have an existing `/v2/repos/lotto` endpoint that returns a list of lottery factor contributors that can be used to populate this chart

```
{
  "all_contribs": [
    {
      "contributor": "nickytonline",
      "count": 34,
      "percent_of_total": 0.3269230769230769,
      "lotto_factor": "high"
    },
    {
      "contributor": "brandonroberts",
      "count": 29,
      "percent_of_total": 0.27884615384615385,
      "lotto_factor": "high"
    },
    ...
  ],
  "all_lotto_factor": "high"
}
```

## The Barchart
The goal is for users to put in perspective the contributions of the lottery factor contributors vs the rest of the community.
To achieve this, the metric for the barchart should be the `percentage_of_total` field on the endpoint, and the chart should:
- Show the contributors that make up to 80% of the contributions, and then the rest of the community grouped

BUT, if there's more than 5 contributors that have made 80% of the contributions:
- Show only the top 5 contributors + the rest of the community grouped

The reason behind this is to avoid crowding the chart too much. 

Two examples of what the bar chart would look like for two different repositories (see color code in next section):
image

### Color Code
image

## The insight subtitle

image

zeucapua:

Closes #3341 

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [ ] Tier 4

### DIFF
--- a/components/Repositories/LotteryFactorChart.tsx
+++ b/components/Repositories/LotteryFactorChart.tsx
@@ -1,0 +1,19 @@
+import { HiOutlineTicket } from "react-icons/hi";
+import Card from "components/atoms/Card/card";
+import Pill from "components/atoms/Pill/pill";
+
+export default function LotteryFactorChart() {
+  return (
+    <Card className="flex flex-col gap-8 w-full h-full items-center p-4">
+      <section className="flex flex-col lg:flex-row w-full items-start lg:items-center gap-4 lg:justify-between px-4 lg:px-8">
+        <header className="flex w-full justify-between items-center">
+          <div className="flex gap-2 items-center">
+            <HiOutlineTicket className="text-xl" />
+            <h3 className="text-sm font-semibold md:text-lg text-slate-800">Lottery Factor</h3>
+          </div>
+          <Pill text="High" color="red" size="base" />
+        </header>
+      </section>
+    </Card>
+  );
+}
--- a/pages/s/[org]/[repo]/index.tsx
+++ b/pages/s/[org]/[repo]/index.tsx
@@ -25,6 +25,7 @@ import ContributorsChart from "components/Graphs/ContributorsChart";
 import { useMediaQuery } from "lib/hooks/useMediaQuery";
 import IssuesChart from "components/Graphs/IssuesChart";
 import PRChart from "components/Graphs/PRChart";
+import LotteryFactorChart from "components/Repositories/LotteryFactorChart";
 
 const AddToWorkspaceModal = dynamic(() => import("components/Repositories/AddToWorkspaceModal"), {
   ssr: false,
@@ -197,6 +198,7 @@ export default function RepoPage({ repoData, ogImageUrl }: RepoPageProps) {
             </div>
           </div>
           <ClientOnly>
+            <LotteryFactorChart />
             <ContributorsChart
               stats={contributorStats}
               range={range}


isabensusan:
I'll review it further, but just collecting on feedback from @bdougie -- Let's add an explainer tooltip to the title to explain what lottery factor is

bdougie:
Note for @jpmcb, we talked IRL. But should filter bots.

isabensusan:
One small comment is that if there hasn't been any commits for the time range selected, it breaks:
image

I'm designing both an empty state and the tooltip rn so we can work them in, put the chart in the layout and ship  :)

components/Repositories/LotteryFactorChart.stories.tsx
```diff
@@ -0,0 +1,124 @@
+import { Meta, StoryObj } from "@storybook/react";
```
nickytonline, line 1: For the Storybook story, show the different lotto pills.

CleanShot 2024-05-10 at 11 16 07

nickytonline (changes requested):

components/Repositories/LotteryFactorChart.tsx
```diff
@@ -0,0 +1,210 @@
...
+        ? 0
+        : Number(
+            (
+              lotteryFactor.all_contribs
+                .slice(0, 4)
+                .reduce((prev, contributor) => prev + contributor.percent_of_total, 0) * 100
+            ).toPrecision(1)
+          ),
+    [lotteryFactor]
+  );
```
nickytonline, line 31: Since only `lotteryFactor` is in the dependency array, that means error or isLoading will always result in 0 because in both those cases it will be `!lotteryFactor`. Is the `isLoading ||` necessary then?

components/Repositories/LotteryFactorChart.tsx
```diff
@@ -0,0 +1,210 @@
...
+                  setHovered(undefined);
+                }}
+              />
+            );
+          })}
+        </div>
+      </section>
+
+      {!lotteryFactor ? (
+        <p>Loading...</p>
```
nickytonline, line 156: Do we want a better loading state @isabensusan?

components/molecules/DevProfile/dev-profile.tsx
```diff
@@ -29,9 +30,7 @@ const DevProfile = ({ username, hasBorder }: DevProfileProps) => {
         <AvatarHoverCard contributor={username} repositories={[]} size="small" />
       </div>
       <div>
-        <h1 className="text-light-slate-12">
-          {username && username.length > 10 ? truncateString(username, 11) : username}
-        </h1>
+        <h1 className="text-light-slate-12">{username && truncate ? truncateString(username, 11) : username}</h1>
```
nickytonline, line 33: Use CSS for this not JS.

next-types.d.ts
```diff
@@ -605,3 +605,15 @@ interface DbWorkspacesReposStats {
...
+  contributor: string;
+  count: number;
+  percent_of_total: number;
+  lotto_factor: "very-high" | "high" | "moderate" | "low";
+}
+
+interface RepositoryLottoFactor {
+  all_contribs: ContributorLottoFactor[];
+  all_lotto_factor: "very-high" | "high" | "moderate" | "low";
+}
```
nickytonline, line 619: Refactor (blocking): No need to duplicate the type.

```suggestion
type LottoFactor = "very-high" | "high" | "moderate" | "low"

interface ContributorLottoFactor {
  contributor: string;
  count: number;
  percent_of_total: number;
  lotto_factor: LottoFactor;
}

interface RepositoryLottoFactor {
  all_contribs: ContributorLottoFactor[];
  all_lotto_factor: LottoFactor;
}
```

nickytonline (changes requested):

components/Repositories/LotteryFactorChart.tsx
zeucapua: added `isLoading` to the dependency array

components/Repositories/LotteryFactorChart.tsx
isabensusan: @zeucapua @nickytonline 
can we do something like this?

- skeleton loader component in the top right corner badge
- skeleton loader component in the header, for the text and the bar chart
- skeleton loader component inside each row in the table

image

nickytonline (commented):
Looking great on all screen sizes!

CleanShot 2024-05-10 at 16 05 14

CleanShot 2024-05-10 at 16 05 49

CleanShot 2024-05-10 at 16 06 12

components/Repositories/LotteryFactorChart.tsx
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    return (
+      <p className="text-slate-500">
+        The top <span className="font-semibold text-black">{count}</span> contributor{count > 1 && "s"} of this
+        repository have made <span className="font-semibold text-black">{percentage}%</span> of all commits in the past{" "}
+        <span className="font-semibold text-black">{range}</span> days.
+      </p>
+    );
+  }, [isLoading, lotteryFactor, range]);
```
nickytonline, line 78: `useMemo` seems overkill here. Maybe for the `topFourContributors` calculation.

nickytonline:
> @zeucapua here's the empty state for this chart as well:
> 
> image
> Image by itself: empty-state

Can we add an empty state example to the Storybook stories as well?

nickytonline (changes requested):
Thanks for making the updates @zeucapua.

One last thing. The loading state doesn't span the whole chart for contributor rows.

CleanShot 2024-05-13 at 12 24 41

nickytonline (approved):
Thanks for making the changes @zeucapua!

🚢

isabensusan (approved):
Looks good to me!

PR #3364, opened 2024-05-10
fix: now opening workspace dropdown puts focus on search input

### LINKED ISSUE #3362: Bug: Opening Workspace dropdown does not put focus on input
When you click or open the workspace dropdown via keyboard, the search input does not receive focus. This is annoying as a mouse user, but unusable as a keyboard user or some other form of assitive technology.

CleanShot 2024-05-09 at 19 37 46

CleanShot 2024-05-09 at 19 40 32

nickytonline:
## Description

Now when you open the workspace dropdown, focus is given to the search input making it more accessible and less annoying to use.

Fixes #3362

**Before**

CleanShot 2024-05-10 at 10 33 48

**After**

CleanShot 2024-05-10 at 10 32 34

## Steps to QA

1. In the app sidebar, click on the workspace dropdown.
2. Notice the search items input automatically has focus.

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


components/atoms/Select/single-select.tsx
```diff
@@ -35,7 +35,14 @@ const SingleSelect = ({
 
   useEffect(() => {
     if (isOpen && isSearchable) {
-      inputRef.current?.focus();
+      // TODO: Revisit this. There's a timing issue where the input's ref is set when the dropdown is opened
+      // but it still can't receive focus. This is a workaround to focus the input after
+      // the dropdown is opened.
+      setTimeout(() => {
```
nickytonline, line 41: As mentioned this is a workaround. I could dig further, but I wasted a lot of time on this already. Basically the `inputRef.current` is non-null but not when `isOpen` and `isSearchable` are true. I tried some things including the non-recommended putting `inputRef.current` in the dependency array but reverted one, because it's not recommmended and can cause unnecessary rerenders, but two, it didn't work. 🙃

Maybe there is something else going on wit the Command* components that I'm unaware of, but for now, I think this workaround it fine.

isabensusan (approved):
perfect-mean-girls-gif

jpmcb (approved):

Issue #3353, opened 2024-05-09
POC: Explore the Vercel AI SDK


nickytonline:
We're currently streaming responses for StarSearch that return markdown which we transform with React Markdown. This works fine, but we're limited to plain HTML tags.

We should explore the [Vercel AI SDK](https://sdk.vercel.ai/docs/introduction) to bring richer UI experiences for StarSearch.

This video is a great watch: https://www.youtube.com/watch?v=br2d_ha7alw

Also, @bdougie shared this AI Chatbot example, https://chat.vercel.ai/

Source code is available here, https://github.com/vercel/ai-chatbot.

nickytonline:
I believe we'll need to upgrade to Next.js 14, #2021, and the AI SDK requires the app router, so if we go ahead with this, we could start with StarSearch for using the App router.

bdougie:
After a quick look, they use the tools/functions pattern to know when to render a component instead of text. It's pretty elegant and makes a ton of sense. 

https://github.com/vercel/ai-chatbot/blob/095550d4dca22dc506cdbda815cab94cfe8fbe74/lib/chat/actions.tsx#L190-L380

nickytonline:
From @jpmcb, https://vercel.com/guides/openai-function-calling

jpmcb:
## Proposal - send `metadata` in the observable to inform the client what components to render

It seems the Vercel Next AI SDK assumes you just drop in an OpenAI token and it will call a GPT with some tools and prompts you right on the server side of a Next app to then render some components. Example:

https://github.com/vercel/ai-chatbot/blob/095550d4dca22dc506cdbda815cab94cfe8fbe74/lib/chat/actions.tsx#L151

This would work really well for an application that primiarily uses server side rendering for calling out to an AI service. But would make having more heavy lifting (like putting together a cosine similarity search for RAG or calculating data based on returned database results) much more difficult. This would also make having several AI agents that have different tasks managed by the API much more difficult to pull off, especially if we integrate more Azure services in the future.

I propose that we keep the current paradigm (where the API does all the heavy lifting, manages AI agents, has it's OpenAI tools, configured Azure services, etc.) but we include an additional field in the observable:

```
metadata: { json }
```

which would denote _which_ OpenAI tools were called so the client can know which components to hit the API with to then render. This way, we're not grid locking us into using server side rendering for components and we keep the flexibility of using an observable sent to the client.

[...]

PR #3367, opened 2024-05-11
feat: Skunksworks - show consuming the "event" from Star-Search SSE

jpmcb:
## Description

This is a skunkswork followup to: https://github.com/open-sauced/api/pull/787

which shows how we can consume the "event" type from a SSE (in this case, the function call) and it's associated payload.

Screenshot 2024-05-10 at 5 45 53 PM

---

This could then be taken to render some components on the client side depending on the function call the model decided to use. This more or less replicates how Vercel's AI SDK works (but without the server side rendering and heavy lifting calling OpenAI happening in the SSE code).

Example, something like:

```js
switch (incomingFunctionCall) {

  case (functionCallEnum.RenderLotteryFactor):
    // render the lottery factor graph with the given json dump from the SSE

  case (functionCallEnum.RenderUser):
    // render the user card for a given user
    // etc. etc.

  break
}
```

---

In this example ^ I asked a question about Brandon and got back from the server the "searchPrsByAuthor" function call that the model decided to make. In the "data" for that event type, we also get parsable JSON that can be used.

The really tricky thing here is the non-deterministic nature of the model:

### 1. LLM non-determinism

We can't always be sure the model will use the functions properly (for example, sometimes it doesn't include a required argument in the function call parameters and it has to go back and try again, which, on the second attempt, it usually gets right). For example: 

```
event: function_call
id: 1
data: {"name":"searchPrsByAuthor","arguments":"{\"question\": \"What recent pull requests has @brandonroberts been involved in?\"}"}

event: function_call
id: 2
data: {"name":"searchIssuesByAuthor","arguments":"{\"question\": \"What recent issues has @brandonroberts been involved in?\"}"}

event: function_call
id: 3
data: {"name":"searchPrsByAuthor","arguments":"{\"question\": \"What recent pull requests has @brandonroberts been involved in?\", \"author\": \"brandonroberts\"}"}

event: function_call
id: 4
data: {"name":"searchIssuesByAuthor","arguments":"{\"question\": \"What recent issues has @brandonroberts been involved in?\", \"author\": \"brandonroberts\"}"}
```

Here, in this example, it tried to use "searchPrsByAuthor" and "searchIssuesByAuthor" _without_ the required "author" field in the parameters. Thankfully, the Zod strongly typed parameters made it go back and try again. And on the second attempt it got it. 

We can probably correct for alot of this by ensuring that the client here always has very strongly typed parameter classes that are used to insure the incoming json is correct. If not, we simply throw it out and don't render something and let it say what it'll say. 

### 2. Using "special" tools/functions for rendered components

On the API side, we might consider instructing the model to be aware of some special tools/functions it has access to that get used when it thinks it should render some components on a client that's connected. This, again, _delves_ into the realm of prompt-engineering so we should find one or two really good use cases to prove out the pattern. Maybe renderding the lottery factor table would be a good example. We instruct the model that it has a "renderLotteryFactorTable" function it can call when there are queries about lottery factor for specific repos. That would return a payload like:

```
event: function_call
id: 1
data: {"name":"renderLotteryFactorTable", "arguments":"{\"repo\": \"kubernetes/kubernetes\"}"}
```

which could be parsed and using the `kubernetes/kubernetes` repo name, we render that table.

Related to #3353 

## Steps to QA

You'll need a dev server from the API skunkworks PR mentioned early.

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/pages/star-search/index.tsx
+++ b/pages/star-search/index.tsx
@@ -167,42 +167,52 @@ export default function StarSearchPage({ userId, bearerToken, ogImageUrl }: Star
         return;
       }
       const values = value.split("\n");
-      values
-        .filter((v) => v.startsWith("data:"))
-        .forEach((v) => {
-          /*
-           * regex for capturing star-search stream SSEs:
-           * data:\s?(?<result>.*)
-           *
-           * The aim of this regex is to capture all characters coming from
-           * the star-search server side events while also preserving the
-           * empty "data:" frames that may come through (which are newlines).
-           *
-           * 'data:' - matches the "data:" characters explicitly.
-           * '\s'    - matches any whitespace that follows the data. In most cases, this is a single space ' '.
-           * '?'     - matches the previous whitespace token zero or one times. Aka, is optional.
-           *
-           * '(?<result>.*)' - optional named capture group "result".
-           *    ├────── '?'          - capture group is optional.
-           *    ├────── '<result>'   - capture group is named "result".
-           *    └────── '.*'         - matches any characters (including zero characters) after the "data:\s?" segment.
-           *                           this is in service of also capturing empty strings as newlines.
-           */
+      values.forEach((v) => {
+        if (v.startsWith("event:")) {
+          // eslint-disable-next-line no-console
+          console.log("event type", v.substring("event: ".length).trim());
+        } else if (v.startsWith("data: ")) {
+          try {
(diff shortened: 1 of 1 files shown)


jpmcb:
Closing as this is surpassed by @nickytonline's excellent work in https://github.com/open-sauced/app/pull/3394

PR #3415, opened 2024-05-17
fix: fixed rendering of lottery chart when less than 4 contributors

### LINKED ISSUE #3402: Bug: Fix rendering of LotteryGraphChart when no data available
### Suggested solution

Fix rendering of LotteryGraphChart when no data available.

nickytonline:
## Description

Fixes the lottery chart when there are less than 4 contributors. Prior to the fix you'd see some broken avatar images and other contributors being consider as a contributor.

Currently if no contributions have been made, it shows as a low lottery factor. This will change to high once @jpmcb's work in the API is merged (see https://github.com/open-sauced/api/issues/808).

Fixes #3402

The numbers might look a little off for one, two, and three contributors but that's because it's not real data.

CleanShot 2024-05-17 at 11 10 59

CleanShot 2024-05-17 at 11 24 55

CleanShot 2024-05-17 at 11 31 52

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/components/Repositories/LotteryFactorChart.stories.tsx
+++ b/components/Repositories/LotteryFactorChart.stories.tsx
@@ -79,6 +79,18 @@ export const AllLottoFactorLow: Story = {
   },
 };
 
+export const NoContributors: Story = {
+  args: {
+    lotteryFactor: {
+      all_contribs: [],
+      all_lotto_factor: "low",
+    },
+    range: 30,
+    isLoading: false,
+    error: undefined,
+  },
+};
+
 export const OneContributor: Story = {
   args: {
     lotteryFactor: {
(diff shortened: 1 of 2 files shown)


BekahHW:
I wonder if we should have a different design for projects with no contributors in the last 30 days:

image

nickytonline:
> deploy-preview-3415--oss-insights.netlify.app

We have a not enough data screen that seems to be used for when it errors. I think we can just show this? But instead of that message we say no activity or something.

CleanShot 2024-05-17 at 10 27 06

nickytonline:
Do we know of some repos with less than 4 contributors? I just want to make sure it's calculating correctly. I think it is, but in Storybook it's made up data which is what threw me off initially.

BekahHW:
@nickytonline pretty sure my blog has less than four: https://github.com/BekahHW/BekahHW.github.io

jpmcb (approved):

PR #3424, opened 2024-05-18
feat: Use StarSearchPayload from API

jpmcb:
## Description

This is a followup to https://github.com/open-sauced/api/pull/818 which introduces a well defined, strongly typed object for sending in the StarSearch `data:` stream. This should make parsing the incoming content much, _much_ easier. This method is heavily inspired by ChatGPT and the data object they send back in their stream.

This also introduces the remark github style markdown plugin. I'm not sure if it's making a difference, but maybe worth playing around with.

Followup to: https://github.com/open-sauced/api/pull/818

⚠️ The deploy preview will not work until the API changes ^go in.

Overall, this seems to fix alot of the messy line breaking that was going on:

Screenshot 2024-05-18 at 4 59 34 PM

## Steps to QA

1. Run the API locally (or wait for API changes to go in).
2. Run app locally, check the /star-search route

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


pages/star-search/index.tsx
```diff
@@ -634,7 +633,11 @@ function Chatbox({ message, userId }: { message: StarSearchChat; userId?: number
   if (typeof message.content == "string") {
     // Breaking all words so that the rendered markdown doesn't overflow the container
     // in certain cases where the content is a long string.
-    content = <Markdown className="break-all">{message.content}</Markdown>;
+    content = (
+      <Markdown remarkPlugins={[remarkGfm]} className="break-all">
```
zeucapua, line 637: Need to change `break-all` to `break-words` to fix overflow that is seen below: 
image

zeucapua (changes requested):
Looks good! Just needs a fix on the text break.

jpmcb:
> The list markdown isn't rendering correctly:

Yeah, I was hoping the github flavored markdown plugin would fix this :( but it's till messed up. Any ideas?

pages/star-search/index.tsx
```diff
@@ -634,7 +633,11 @@ function Chatbox({ message, userId }: { message: StarSearchChat; userId?: number
   if (typeof message.content == "string") {
     // Breaking all words so that the rendered markdown doesn't overflow the container
     // in certain cases where the content is a long string.
-    content = <Markdown className="break-all">{message.content}</Markdown>;
+    content = (
+      <Markdown remarkPlugins={[remarkGfm]} className="break-words">
```
zeucapua, line 637: It renders the list to the correct HTML, just gotta style it with Tailwind. Just add `prose` to the className for `Markdown`

image

zeucapua (changes requested):

pages/star-search/index.tsx
```diff
@@ -365,33 +357,36 @@ export default function StarSearchPage({ userId, bearerToken, ogImageUrl }: Star
...
+            const { result } = matched.groups;
+
+            // deserialize the json dump from the payload
+            const payload = JSON.parse(result) as StarSearchPayload;
+
             // function_call means we're loading a widget definition for the enriched UI
-            if (eventType === "function_call") {
-              jsonContent = JSON.parse(result);
+            if (payload.content.type === "function_call") {
+              jsonContent = JSON.parse(payload.content.parts[0]);
```
brandonroberts, line 369: Is this guaranteed to have at least one item in the array?

jpmcb: Yes: but probably better to add a check 👍🏼

pages/star-search/index.tsx
jpmcb: Oh yeah, this looks so much better 👏🏼 👏🏼 good find @zeucapua 🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅🏅

jpmcb:
`prose` seemed to be the silver bullet we needed to get these to render nicely - huge shoutout to @zeucapua for discovering that!!! 👏🏼 

Screenshot 2024-05-20 at 2 53 10 PM

Also added some additional comments and a check that `result.content.parts` has some actual stuff.

pages/star-search/index.tsx
```diff
@@ -301,97 +300,100 @@ export default function StarSearchPage({ userId, bearerToken, ogImageUrl }: Star
...
-           *    └────── '.*'         - matches any characters (including zero characters) after the "data:\s?" segment.
-           *                           this is in service of also capturing empty strings as newlines.
+           * Example: data: { id: "abc123" }
+           * - Captures the 'data: ' (including the space)
+           * - The named capture group "result" gets the entire JSON object "{ id: \"abc123\" }"
+           *   as a string that can be parsed to an object.
            */
 
-          const matched = v.match(/data:\s?(?<result>.*)/);
+          const matched = v.match(/data:\s(?<result>.*)/);
```
jpmcb, line 356: I made the regex _just abit_ more strict: now, there has to be a space between `data:` and the json object flowing in. We can always expect the same result now from the API: the `data` key, a `:`, a space ` `, and then the json object. So, for example, `data: { id: "abc123" }` will get captured. But `data:` or `data:{ ... }` will be rejected.

pages/star-search/index.tsx
```diff
@@ -301,97 +300,100 @@ export default function StarSearchPage({ userId, bearerToken, ogImageUrl }: Star
...
+            let jsonContent: WidgetDefinition;
+            const { result } = matched.groups;
+
+            // deserialize the json dump from the payload
+            const payload = JSON.parse(result) as StarSearchPayload;
+
+            // skip over cases where the payload is somehow malformed or missing content altogether
+            if (!payload || !payload.content || payload.content.parts.length === 0) {
+              return;
+            }
```
jpmcb, line 372: Small check here: maybe pointless since `JSON.parse(result) as StarSearchPayload` _I think_ will throw if it can't parse all the non-optional pieces of the json string.

brandonroberts: JSON.parse will throw if the JSON is invalid, but won't throw against checking for specific keys or the shape of the content. The type is good is though for any following consumers

brandonroberts (approved):
Nice work! 🤝

jpmcb:
Looking good with the chat bubbles!! cc @nickytonline 

Screenshot 2024-05-21 at 8 29 21 AM

would love to get this in a release today so we can play around with everything before thursday!

pages/star-search/index.tsx
```diff
@@ -302,97 +302,100 @@ export default function StarSearchPage({ userId, bearerToken, ogImageUrl }: Star
...
-
           try {
+            let jsonContent: WidgetDefinition;
+            const { result } = matched.groups;
+
+            // deserialize the json dump from the payload
+            const payload = JSON.parse(result) as StarSearchPayload;
+
+            // skip over cases where the payload is somehow malformed or missing content altogether
+            if (!payload || !payload.content || payload.content.parts.length === 0) {
```
nickytonline, line 372: consider logging to Sentry here to surface the invalid payload.

next-types.d.ts
```diff
@@ -619,3 +619,27 @@ interface RepositoryLottoFactor {
...
   all_lotto_factor: LottoFactor;
 }
+
+// sourced from open-sauced/api
+
+type StarSearchEvent = "content" | "final" | "function_call";
+type StarSearchPayloadStatus = "in_progress" | "done";
+
+interface StarSearchContent {
+  type: string;
```
nickytonline, line 629: Is the type here `StarSearchEvent` or is this as narrow as the type can be?

pages/star-search/index.tsx
nickytonline: always one space or can it be `\s+`? I'd assume just `\s` since the API is handling the formatting?

nickytonline (approved):
Nothing blocking, just some comments. Great work @jpmcb!

🚢

pages/star-search/index.tsx
jpmcb: Yes, correct, the payload from the API should always be `data: { ... }` with one space and we should reject anything that is malformed. It's probably a moot point to make the regex more strict, but it at least makes the client have to do _alot_ less heavy lifting since the API is doing it now

pages/star-search/index.tsx
```diff
@@ -302,97 +302,103 @@ export default function StarSearchPage({ userId, bearerToken, ogImageUrl }: Star
...
+              Sentry.captureException(
+                new Error(`Parsed and rejected malformed JSON for StarSearch. JSON payload: ${v}`)
+              );
```
jpmcb, line 375: Sends Sentry an error if we get some weird malformed JSON for some reason.

nickytonline:
> Looking good with the chat bubbles!! cc @nickytonline
> 
> Screenshot 2024-05-21 at 8 29 21 AM
> 
> would love to get this in a release today so we can play around with everything before thursday!

Not sure why you're loader is blue. 🤔 Here it is on beta.

CleanShot 2024-05-21 at 11 40 54

nickytonline:
> deploy-preview-3424--oss-insights.netlify.app

Not sure what's going on your local, but on your deploy preview the loader animation is orange. 🙃

PR #3433, opened 2024-05-21
fix: add support for clipboard copy in Safari and use utility function

### LINKED ISSUE #3385: Bug: Share buttons do not work on mobile
Trying to click the share buttons and get no interaction on mobile.

Profile

image

Repo

image

Repo Insights

image

brandonroberts:
## Description

- Fixes a bug with copying text to the clipboard in Safari-based browsers
- Adds a utility function to copy text to the clipboard that works in Chrome and Safari

Closes #3385 

## Steps to QA

1. Go to any page with a share button (repo page, workspace, etc) in Safari on Desktop or iOS
2. Click the `Share` button
3. Note the short URL is copied to the clipboard and the toast message is displayed

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/lib/utils/write-to-clipboard.ts
+++ b/lib/utils/write-to-clipboard.ts
@@ -1,0 +1,3 @@
+export async function writeToClipboard(url: string) {
+  setTimeout(async () => await navigator.clipboard.writeText(url));
+}
--- a/components/ListHeader/list-header.tsx
+++ b/components/ListHeader/list-header.tsx
@@ -42,7 +42,7 @@ const ListHeader = ({
 
     try {
       const shortUrl = await shortenUrl(url);
-      await writeToClipboard(shortUrl);
+      writeToClipboard(shortUrl);
       toast({ description: "Copied to clipboard", variant: "success" });
     } catch (error) {
       // eslint-disable-next-line no-console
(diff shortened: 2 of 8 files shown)


lib/utils/write-to-clipboard.ts
```diff
@@ -0,0 +1,3 @@
+export async function writeToClipboard(url: string) {
+  setTimeout(async () => await navigator.clipboard.writeText(url));
```
nickytonline, line 2: Didn't realize it was a timing issue. I thought it was a security issue.

lib/utils/write-to-clipboard.ts
```diff
@@ -0,0 +1,3 @@
+export async function writeToClipboard(url: string) {
```
nickytonline, line 1: The function no longer needs to be async as it's the `setTimeout` callback that handles the async function call now. The `await writeToClipboard(shortUrl);` can be `writeToClipboard(shortUrl);` now.

```suggestion
export function writeToClipboard(url: string) {
```

lib/utils/write-to-clipboard.ts
brandonroberts: There's different ways to resolve it. You can use `ClipboardItem` but this breaks in Chrome. I got the simpler solution from here

https://stackoverflow.com/questions/66312944/javascript-clipboard-api-write-does-not-work-in-safari#comment126772784_68241516

nickytonline (approved):
:shipit:

jpmcb (approved):
📎 📎 📎

PR #3492, opened 2024-05-28
fix: redirect to 404 if user not found

### LINKED ISSUE #3408: 404 when users don't exist
The search kind of encourages this but when a user doesn't exist we render the page.

We should really try to 404 this.

I.e. Perhaps we can do something similar for accounts that don't exist. Seems we have a great 404, but an aggressive router.

This should 404 https://app.opensauced.pizza/user/user8675

Similar with how organizations appear as users.

image

zeucapua:
## Description

Redirect from the `/user` page to 404 if user doesn't exist.

Closes #3408 

## Steps to QA

1. Go to deploy preview and go to `/user/user8675`
2. See 404 page

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


pages/user/[username]/index.tsx
```diff
@@ -106,38 +106,25 @@ Contributor.PageLayout = ProfileLayout;
...
 
-      return (await req.json()) as DbUser;
-    } catch (e) {
-      // eslint-disable-next-line no-console
-      console.error(e);
-      return null;
-    }
+  if (!req.ok) {
+    return {
+      redirect: { destination: "/404" },
```
Comment (brandonroberts) [inline pages/user/[username]/index.tsx:118]: Next has a built-in property we can return for a 404

```suggestion
      notFound: true,
```

brandonroberts (changes requested):

pages/user/[username]/index.tsx
```diff
@@ -106,38 +106,25 @@ Contributor.PageLayout = ProfileLayout;
...
-        headers: {
-          accept: "application/json",
-        },
-      });
-
-      if (!req.ok) {
-        return null;
-      }
+  const { username } = context.params!;
+  const req = await fetch(`${process.env.NEXT_PUBLIC_API_URL}/users/${username}`, {
```
Comment (nickytonline) [inline pages/user/[username]/index.tsx:110]: you can try the following for the server-side request forgery warning. The `username` should be validated as well to avoid something like `username = `../user/delete`.

```suggestion
  const req = await fetch(`${new URL("/users/${username}", process.env.NEXT_PUBLIC_API_URL})`, {
```

brandonroberts (approved):

pages/user/[username]/index.tsx
Comment (jpmcb) [reply pages/user/[username]/index.tsx:110]: There's also the URL param validators functions I wrote in: https://github.com/open-sauced/app/pull/3467

```suggestion
  if (!isValidUrlSlug(username)) {
    throw new Error("Invalid input");
  }
  
  const req = await fetch(`${process.env.NEXT_PUBLIC_API_URL}/users/${username}`, {
```

which should in theory always error if an invalid slug (like trying to escape a URL with `;`) is provided

bdougie (approved):

pages/user/[username]/index.tsx
Comment (jpmcb) [reply pages/user/[username]/index.tsx:116]: Looks like it's not complaining about the request forgery anymore 👍🏼 😎 👍🏼

jpmcb (approved):

PR #3531, opened 2024-06-05
fix: set `/star-search` user avatar after logging in

### LINKED ISSUE #3523: Bug: Avatar link broken when logged out user logs in on StarSearch page
If you're logged in, your avatar for StarSearch always appears.

CleanShot 2024-06-05 at 07 37 06

However, if you start from a logged out experience and log in, your avatar is broken.

CleanShot 2024-06-05 at 07 42 25

This wasn't an issue before, but now that we let users access `/star-search` and prompt them to log in if not logged in it's broken because we're retrieving the `userId` via `getServerSideProps` which only renders those props on the initial page load. The `userId` prop should be promoted to a variable within the component to handle the logged out then logged in scenario.

zeucapua:
## Description

Added a `useMemo` to set the user ID based on the session. This fixes the issue where it doesn't render any image if the user starts logged out in the StarSearch page then logs in when prompted.

Closes #3523 

https://github.com/open-sauced/app/assets/20603494/dffdf510-951e-4f78-a045-0161fe8893d9

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/pages/star-search/index.tsx
+++ b/pages/star-search/index.tsx
@@ -1,7 +1,7 @@
 import { GetServerSidePropsContext } from "next";
 import { createPagesServerClient } from "@supabase/auth-helpers-nextjs";
 import { MdOutlineSubdirectoryArrowRight } from "react-icons/md";
-import { Fragment, useEffect, useRef, useState } from "react";
+import { Fragment, useEffect, useMemo, useRef, useState } from "react";
 
 import Image from "next/image";
 import Markdown from "react-markdown";
@@ -29,6 +29,7 @@ import { StarSearchLoader } from "components/StarSearch/StarSearchLoader";
 import StarSearchLoginModal from "components/StarSearch/LoginModal";
 import useSupabaseAuth from "lib/hooks/useSupabaseAuth";
 import AvatarHoverCard from "components/atoms/Avatar/avatar-hover-card";
+import useSession from "lib/hooks/useSession";
 
 export interface WidgetDefinition {
   name: string;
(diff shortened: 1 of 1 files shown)


pages/star-search/index.tsx
```diff
@@ -189,7 +190,9 @@ export default function StarSearchPage({ userId, ogImageUrl }: StarSearchPagePro
   const scrollRef = useRef<HTMLDivElement>(null);
   const { feedback, prompt } = useStarSearchFeedback();
   const { toast } = useToast();
+  const { session } = useSession(true);
   const { sessionToken: bearerToken } = useSupabaseAuth();
+  const userId = useMemo(() => (session ? session.id : initialUserId), [session]);
```
nickytonline, line 195: There will only ever be the currently logged in user here. `useMemo` seems unnecessary as this is not an expensive calculation.

pages/star-search/index.tsx
```diff
@@ -178,7 +179,7 @@ function StarSearchWidget({ widgetDefinition }: { widgetDefinition: WidgetDefini
   );
 }
 
-export default function StarSearchPage({ userId, ogImageUrl }: StarSearchPageProps) {
+export default function StarSearchPage({ userId: initialUserId, ogImageUrl }: StarSearchPageProps) {
```
nickytonline, line 182: No need to have an `initialUserId`. We can remove the server-side prop for the userId and manage it only on the client-side.

nickytonline (changes requested):

brandonroberts (approved):

nickytonline (approved):
Works great! 🚢

PR #3569, opened 2024-06-17
fix: update user search to fallback to GitHub API on error and zero results

brandonroberts:
## Description

- Updates the user search utility function to search the GitHub API under separate conditions
  - Our primary API fails
  - Our primary API returns 0 results
- If the GitHub API fails then we return `false` which indicates no results were found in either. 

## Steps to QA

1. Go to deploy preview
2. Open the network panel in browser devtools
3. Search for a user that's not in our data
4. See two requests made, one to our API, and one to the GitHub api
5. See results from the searches

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


nickytonline (approved):
🚢

jpmcb (approved):

PR #3563, opened 2024-06-13
feat: enabled loading an existing conversation in StarSearch

nickytonline:
## Description

Now you can share an existing thread (conversation in StarSearch). We had a rudimentary version of this previously (see #3324), but it was just sharing the prompt and running it.

These shared conversations are the full conversation that gets replayed and does not hit our AI backend at all.

If you are not logged in, you will only be able to replay it once (unless the page is refreshed in the browser) and any other actions will require you to log in.

One thing to note is if you generate a shared URL, it will generate for beta.app.opensauced.pizza on beta and deploy previews, and one in prod, will be prod links. All that said, you'll just need to replace the beta base URL with the deployment preview URL.

TODO:
- fix flash of content when loading a conversation
~- fix OG image for a shareable StarSearch conversation~
~- fix Server-side request forgery warnings~

**Note: This PR adds [Valibot](https://app.opensauced.pizza/s/fabian-hiller/valibot) to the project. It is MIT licensed and a small package (1kb). It's used for some validation in this PR and I plan to use it elsewhere in the application long term.**

## Related Tickets & Documents

Closes #3551 

## Steps to QA

Here's a few share links you can try:

* [Recent Contributions of Brandon Roberts to Analogjs Repository](https://deploy-preview-3563--oss-insights.netlify.app/star-search?share_id=900686cc-8926-47fd-a1d6-0e19da967f48)

Associated OG image, https://deploy-preview-3563--oss-insights.netlify.app/og-images/star-search/?share_id=900686cc-8926-47fd-a1d6-0e19da967f48

image

* [Dependency Assessment of Key Contributors in remix-run/react-router Project](https://deploy-preview-3563--oss-insights.netlify.app/star-search?share_id=ff9b26b7-64e7-460b-9d33-252543bee24d)

Associated OG image, https://deploy-preview-3563--oss-insights.netlify.app/og-images/star-search/?share_id=ff9b26b7-64e7-460b-9d33-252543bee24d

image

* Note that the [StarSearch page without a shared conversation](https://deploy-preview-3563--oss-insights.netlify.app/star-search/), should load the normal OG image, https://deploy-preview-3563--oss-insights.netlify.app/og-images/star-search/

--

Also try and create a new conversation then try to share it.

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [x] Tier 3
- [ ] Tier 4

PR #3581, opened 2024-06-18
fix: now the title of a thread is set to the initial prompt of a StarSearch conversation

nickytonline:
## Description

Fixes an issue where StarSearch might not have updated the thread title with a summary of the initial StarSearch conversation.

Note that StarSearch will eventually update the title of the thread with the summary. This is a fix in the event that someone shares before that's happened.

Relatest to #3563

**Before**

CleanShot 2024-06-18 at 16 32 50@2x

**After**

CleanShot 2024-06-18 at 16 30 49

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/components/StarSearch/StarSearchChat.tsx
+++ b/components/StarSearch/StarSearchChat.tsx
@@ -455,15 +455,30 @@ export function StarSearchChat({
 
       const payload = await starSearchThreadResponse.json();
       id = payload.id;
-      setChatId(id);
-    }
 
-    try {
-      parseSchema(UuidSchema, id);
-    } catch (error) {
-      captureException(new Error(`Failed to parse UUID for StarSearch. UUID: ${chatId}`, { cause: error }));
-      chatError(true);
-      return;
+      try {
+        parseSchema(UuidSchema, id);
+      } catch (error) {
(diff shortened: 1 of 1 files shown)


components/StarSearch/StarSearchChat.tsx
```diff
@@ -455,15 +455,30 @@ export function StarSearchChat({
...
+        body: JSON.stringify({ title: prompt }),
+        headers: {
+          Accept: "*/*",
+          "Content-Type": "application/json",
+          Authorization: `Bearer ${bearerToken}`,
+        },
+      });
+
+      if (updateStarSearchThreadTitle.status !== 200) {
+        captureException(new Error(`Failed to update StarSearch thread title. UUID: ${id}`));
```
nickytonline, line 478: If it fails updating the title, I'm logging it but letting things continue as we already have the thread ID for the new conversation, and StarSearch will update it eventually with the summary.

brandonroberts (approved):

components/StarSearch/StarSearchChat.tsx
nickytonline: In future, we could pass a title when creating the thread, but for now this works.

zeucapua (approved):
🚀

Issue #3607, opened 2024-06-20
Filter inactive repos when import from an org


bdougie:
When adding repos to a workspace or insight, we have an option to add based on the org. When showing this off today on a call, it was immediately called out that most of the projects are not actively maintained. 

![Screenshot 2024-06-20 at 2 53 23 PM](https://github.com/open-sauced/app/assets/5713670/06d10c3f-d644-4298-9b28-05652f2dc0f9)

## Potential Solutions

1. Provide an option to remove projects that have received no contributions in the last 6 months. 
2. Provide the option to remove repos that are inactive when you edit the workspace. 

example: 
https://oss.fyi/pQnJ4cN

![Screenshot 2024-06-20 at 2 56 02 PM](https://github.com/open-sauced/app/assets/5713670/2ed88517-1039-4206-9532-40bebf8b5b0b)

PR #3665, opened 2024-07-02
feat: add range filter when importing GitHub repos for an org

brandonroberts:
## Description

- Adds a day range filter when importing repos for a GitHub org
- Filters the imported repo list by the range based on the repo `pushed_at` date
- Also fixes an issue with importing large orgs like `microsoft`

Closes https://github.com/open-sauced/app/issues/3607

image

## Steps to QA

1.  Login
2. Go to settings for a workspace
3. Click Add Repositories
4. Click on Import GitHub Organization
5. Select an Org
6. Note the repos is filtered to the date range

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [x] Tier 3
- [ ] Tier 4

### DIFF
--- a/components/Workspaces/TrackedRepoWizard/SelectOrgReposStep.tsx
+++ b/components/Workspaces/TrackedRepoWizard/SelectOrgReposStep.tsx
@@ -88,9 +88,7 @@ export const SelectOrgReposStep = ({
           <Avatar contributor={organization} size="small" />
           <span className="flex align-middle">{organization}</span>
         </div>
-        <div className="flex justify-end">
-          <DayRangePicker onDayRangeChanged={(e) => setRange(Number(e))} />
-        </div>
+        <DayRangePicker onDayRangeChanged={(e) => setRange(Number(e))} />
       </div>
       <p>
         Select the organization repositories updated in the last <span className="font-bold">{range} days</span> that
--- a/components/Workspaces/TrackedRepoWizard/SelectOrgReposStep.tsx
+++ b/components/Workspaces/TrackedRepoWizard/SelectOrgReposStep.tsx
@@ -1,5 +1,7 @@
 import { useEffect, useState } from "react";
+import { useRouter } from "next/router";
 import { Avatar } from "components/atoms/Avatar/avatar-hover-card";
+import { DayRangePicker } from "components/shared/DayRangePicker";
 import { SearchedReposTable } from "../SearchReposTable";
 
 interface SearchByReposStepProps {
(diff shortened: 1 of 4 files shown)


zeucapua (approved):

components/Workspaces/TrackedRepoWizard/SelectOrgReposStep.tsx
```diff
@@ -78,11 +83,19 @@ export const SelectOrgReposStep = ({
...
+          <span className="flex align-middle">{organization}</span>
+        </div>
+        <div className="flex justify-end">
+          <DayRangePicker onDayRangeChanged={(e) => setRange(Number(e))} />
+        </div>
       </div>
-      <p>Select the organization repositories that you want to track.</p>
+      <p>
+        Select the organization repositories updated in the last <span className="font-bold">{range} days</span> that
+        you want to track.
```
jpmcb, line 97: Thinking of an edge case here: is there now no way to import all repos from an org that go beyond being updated in 1 year? I.e., all repos regardless of a filter. For example, with the OpenSauced org, we end up missing all those old repos that we haven't touched in some time.

Which, if this is the intend, then that makes sense 👍🏼

components/Workspaces/TrackedRepoWizard/SelectOrgReposStep.tsx
```diff
@@ -78,11 +83,19 @@ export const SelectOrgReposStep = ({
...
-        <Avatar contributor={organization} size="xsmall" />
-        <span>{organization}</span>
+      <div className="flex w-full justify-between gap-2 font-semibold">
+        <div className="flex gap-2 justify-start align-middle">
+          <Avatar contributor={organization} size="small" />
+          <span className="flex align-middle">{organization}</span>
+        </div>
+        <div className="flex justify-end">
+          <DayRangePicker onDayRangeChanged={(e) => setRange(Number(e))} />
+        </div>
```
nickytonline, line 93: refactor: no need for the div wrapper.

```suggestion
        <DayRangePicker onDayRangeChanged={(e) => setRange(Number(e))} />
```

nickytonline (approved):
One small tweak then good to 🚢

components/Workspaces/TrackedRepoWizard/SelectOrgReposStep.tsx
brandonroberts: Yes, that's correct. Feedback given to @bdougie was that older repos that haven't been updated shouldn't be listed on the org import. I suppose they could go back and copy/paste the other ones in

components/Workspaces/TrackedRepoWizard/SelectOrgReposStep.tsx
brandonroberts: Thanks!

PR #3783, opened 2024-07-25
fix: now pinned table columns only get pinned when embedded StarSearch is closed

### LINKED ISSUE #3773: Bug: Avatar hovercards in embedded StarSearch chat are behind the panel
Avatar hovercards in embedded StarSearch chat are behind the panel that contains StarSearch. This happens because the panel has a `z-index: 1`.

CleanShot 2024-07-25 at 08 40 59

The `z-index: 1` was added because when on workspace activity pages (pull requests and issues), the tables have some fixed columns set via z-index: 1 (recommendation in TanStack docs). Prior to adding the `z-index: 1` to the StarSearch embed panel, the tables would leak through the StarSearch embed panel.

There's a bit of a stacking context/z-index battle going on here. I think either at a `z-index: 1` on the AvatarHovarCard component or using plain CSS in global.css to remove the table z-index when the StarSearch embed panel is open.

nickytonline:
## Description

Now hover cards appear over the StarSearch for Workspaces panel and are no longer hidden.

Fixes #3773

**Before**

CleanShot 2024-07-25 at 19 54 35

**After**

CleanShot 2024-07-25 at 19 54 11

## Steps to QA

1. Test this on the two activity pages of workspaces: pull request and issue pages.
2. Ask StarSearch for workspaces a question about contributors.
3. It should return a response with user links that have the hover avatar beside them.
4. Hover over the avatar.
5. Notice the hover card is on top of everything else.

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/styles/globals.css
+++ b/styles/globals.css
@@ -282,3 +282,12 @@ body:has([aria-hidden="false"][data-star-search-mobile]) {
 body:has([data-star-search-mobile]) .top-nav-container {
   @apply fixed;
 }
+
+/**
+  * The pinned-table-column CSS class is for our tables that have pinned columns.
+  * When embedded StarSearch is open, we want to ensure that the columns unpin so they
+  * don't leak throug the embedded StarSearch panel.
+  */
+body:not(:has([data-star-search-mobile][aria-hidden="false"])) .pinned-table-column {
+  z-index: 1;
+}
--- a/components/StarSearch/StarSearchEmbed.tsx
+++ b/components/StarSearch/StarSearchEmbed.tsx
@@ -106,7 +106,6 @@ export const StarSearchEmbed = ({
         style={{
           top: "var(--top-nav-height)",
           height: "calc(100dvh - var(--top-nav-height))",
-          zIndex: 1,
         }}
       >
         <StarSearchChat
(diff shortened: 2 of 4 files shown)


styles/globals.css
```diff
@@ -282,3 +282,12 @@ body:has([aria-hidden="false"][data-star-search-mobile]) {
 body:has([data-star-search-mobile]) .top-nav-container {
   @apply fixed;
 }
+
+/**
+  * The pinned-table-column CSS class is for our tables that have pinned columns.
+  * When embedded StarSearch is open, we want to ensure that the columns unpin so they
+  * don't leak throug the embedded StarSearch panel.
+  */
+body:not(:has([data-star-search-mobile][aria-hidden="false"])) .pinned-table-column {
```
nickytonline, line 291: This works when not in mobile view as well. This attribute, `data-star-search-mobile` always exists, it's just either `"true"` or `"false"` and the `[aria-hidden="false"]` indicates that the panel is open so we shouldn't pin the table columns.

BekahHW (approved):

brandonroberts (approved):

PR #3868, opened 2024-08-06
feat: now the OSCR has a not calculated state

### LINKED ISSUE #3835: Feature: use `devstats_updated_at` field for OSCR "not calculated yet" state
### Suggested solution

There's a bit of a chicken vs. egg problem when it comes to sorting on the contributor table via the OSCR: we don't always have all the contributor's OSCR's calculated. And while I thought it _might_ be tenable to go and fetch the OSCR for those contributors in real time when the table is queried for, it seems it's unfortunately abit too heavy to go and crunch all those numbers at once. I.e., for large repos with many 10s of contributors, trying to crunch the OSCR for all those contributors and then display a sorted table leads to a pretty bad experience and could take 30 seconds + for the table to load.

---

We should utilize the `devstats_updated_at` field, which will be the zero value timestamp at the start of epoch time:

```json5
{
  "id": 23109390,

  // etc. etc. other user metadata

  "oscr": 0,
  "devstats_updated_at": "1970-01-01 00:00:00+00Z",
}
```

If this date is `1970-01-01 00:00:00+00Z`, which is the zero value for POSIX epoch time, then the client can know that the OSCR has yet to be calculated. The zero value for oscr will always be `0` (and not null) since we want to be conservative when displaying someone's contributor rating.

nickytonline:
## Description

Now the OSCR has a not calculated state.

Blocked by https://github.com/open-sauced/api/issues/1011

Closes #3835

CleanShot 2024-08-06 at 11 15 25

## Steps to QA

This is potentially hard to test as it's uncertain which users have not had their OSCR calculated yet. You could always reset a user's `devstats_updated_at` in the dev database or grab the API locally with a fresh DB.

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


pages/workspaces/[workspaceId]/contributors.tsx
```diff
@@ -68,6 +68,7 @@ export default function WorkspaceContributorsPage({ workspace }: WorkspaceContri
           username: info.contributor.username,
           updated_at: info.contributor.updated_at,
           user_id: info.contributor_id,
+          devstats_updated_at: info.contributor.updated_at,
```
Comment (brandonroberts) [inline pages/workspaces/[workspaceId]/contributors.tsx:71]: Is this supposed to be `info.contributor.devstats_updated_at`?

Comment (nickytonline) [reply pages/workspaces/[workspaceId]/contributors.tsx:71]: oops, yep. Will fix.

Comment (nickytonline) [reply pages/workspaces/[workspaceId]/contributors.tsx:71]: I also have a PR up for the API that will make this PR work.

Comment (nickytonline) [reply pages/workspaces/[workspaceId]/contributors.tsx]: Fixed @brandonroberts!

jpmcb (approved):

components/Tables/ContributorsTable.tsx
```diff
@@ -132,7 +133,13 @@ const defaultColumns = ({ repository, isLoggedIn }: { repository?: string; isLog
     ),
     enableSorting: true,
     enableGlobalFilter: false,
-    cell: (info) => <OscrPill rating={info.row.original.oscr ?? 0} hideRating={!isLoggedIn} />,
+    cell: (info) => (
+      <OscrPill
+        rating={info.row.original.oscr}
+        hideRating={!isLoggedIn}
+        calculated={info.row.original.devstats_updated_at !== INITIAL_DEV_STATS_TIMESTAMP}
```
nickytonline, line 140: This is pretty much the logic everywhere below as well.

bdougie (approved):

zeucapua (approved):

PR #3931, opened 2024-08-12
fix: now the back of the dev card no longer peers through when flipped on Safari and Firefox

### LINKED ISSUE #3927: Bug: Dev card back and front overlap when flipped
### Describe the bug

image

### Steps to reproduce

  https://app.opensauced.pizza/u/Kampouse/card 
flip a  card (firefox,safari,mobile)

nickytonline:
## Description

Te dev card back and front no longer overlap when flipped on Safari and Firefox

Fixes #3927

### Before

**Safari**

CleanShot 2024-08-12 at 18 33 31

**Firefox**

CleanShot 2024-08-12 at 18 32 19

**Chromium**

CleanShot 2024-08-12 at 18 48 26

After

**Safari**

CleanShot 2024-08-12 at 18 47 37

**Firefox**

CleanShot 2024-08-12 at 18 46 40

**Chromium**

CleanShot 2024-08-12 at 18 48 26

### Text Selection on Card Flip

I also fixed an issue where the card would get text or images selected within the card.

**Before**

CleanShot 2024-08-12 at 19 22 37

**After**

CleanShot 2024-08-12 at 19 27 48

## Steps to QA

Test the following in a Chromium based browser, Firefox and Safari.

1. Navigate to a dev card page for a user, e.g. https://deploy-preview-3931--oss-insights.netlify.app/u/nickytonline/card
2. Flip the card using the mouse or touch if on mobile.
3. Notice the dev card back and front no longer overlap when flipped

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/components/molecules/DevCard/dev-card.tsx
+++ b/components/molecules/DevCard/dev-card.tsx
@@ -62,7 +62,7 @@ export default function DevCard(props: DevCardProps) {
 
   return (
     <div
-      className="DevCard"
+      className="DevCard select-none"
       style={{
         width: "245px",
         height: "348px",
--- a/components/molecules/DevCard/dev-card.tsx
+++ b/components/molecules/DevCard/dev-card.tsx
@@ -4,6 +4,7 @@ import Tilt from "react-parallax-tilt";
 import { FiGlobe } from "react-icons/fi";
 import { PiCrownSimpleFill } from "react-icons/pi";
 import { HiArrowNarrowRight, HiTrendingDown, HiTrendingUp } from "react-icons/hi";
+import clsx from "clsx";
 import Button from "components/shared/Button/button";
 import openSaucedImg from "img/openSauced-icon.png";
 import { getRelativeDays } from "lib/utils/date-utils";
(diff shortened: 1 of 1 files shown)


components/molecules/DevCard/dev-card.tsx
```diff
@@ -102,7 +103,12 @@ export default function DevCard(props: DevCardProps) {
             height={348}
           />
 
-          <div className="z-10 flex flex-col gap-2 items-center justify-center w-full h-full text-white">
+          <div
+            className={clsx(
+              "z-10 flex flex-col gap-2 items-center justify-center w-full h-full text-white",
+              isFlipped && "invisible"
```
nickytonline, line 109: when it's flipped, set the front of the card to `visibility: hidden`.

zeucapua (approved):

bdougie:
This works but the QA monster in me can't definitely break it if I click fast enough.image

I can get the check to be hidden by cancelling the flip with a click. 

This is much better than what we have but just sharing an edge case.

nickytonline:
> This works but the QA monster in me can't definitely break it if I click fast enough.image
> 
> I can get the check to be hidden by cancelling the flip with a quick tap before it flips fully.
> 
> This is much better than what we have but just sharing an edge case.

We can probably throttle the clicking.

nickytonline:
> This works but the QA monster in me can't definitely break it if I click fast enough.image
> 
> I can get the check to be hidden by cancelling the flip with a quick tap before it flips fully.
> 
> This is much better than what we have but just sharing an edge case.

I'd say ship as is as, but I'll create an issue for that. The component currently only handles clicks which translate to touches on the phone, but we can probably handle your QA gestures. 😅 

I clicked as fast as I could on desktop, and all good though.

bdougie (approved):
Let's move with what we have and follow up with the other things in a future PR.

brandonroberts (approved):

PR #3951, opened 2024-08-14
feat: update onboarding steps to include sharing the DevCard

brandonroberts:
## Description

Updates the onboarding flow to authenticate, set your time zone, and share your dev card.

https://github.com/user-attachments/assets/042748e6-c594-43b2-a5c3-8bbea4110fb9

## Steps to QA

1. Reset your onboarding flag
2. Login
3. Click on `Complete Onboarding` from the nav bar
4. Go through onboarding

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/pages/start.tsx
+++ b/pages/start.tsx
@@ -28,12 +28,10 @@ import useSession from "lib/hooks/useSession";
 import { captureAnalytics } from "lib/utils/analytics";
 
 import useStore from "lib/store";
-import { getInterestOptions } from "lib/utils/getInterestOptions";
 
 import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from "components/atoms/Select/select";
 import { timezones } from "lib/utils/timezones";
 import { useFetchUser } from "lib/hooks/useFetchUser";
-import { LanguageSwitch } from "components/shared/LanguageSwitch/language-switch";
 
 type StepKeys = "1" | "2" | "3";
 
@@ -70,7 +68,7 @@ const LoginStep1: React.FC<LoginStep1Props> = ({ user }) => {
     } else if (onboarded === false && user && providerToken) {
       setQueryParams({ step: "2" } satisfies QueryParams);
     }
-  }, [user, onboarded, providerToken]);
+  }, [user, onboarded, providerToken, router]);
 
   const handleGitHubAuth = async () => {
     // Redirect user to GitHub to authenticate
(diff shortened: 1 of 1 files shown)


pages/start.tsx
```diff
@@ -242,12 +166,12 @@ const LoginStep3: React.FC<LoginStep3Props> = ({ interests, user }) => {
           "Content-Type": "application/json",
           Authorization: `Bearer ${sessionToken}`,
         },
-        body: JSON.stringify({ interests, timezone }),
+        body: JSON.stringify({ interests: [], timezone }),
```
nickytonline, line 169: this is because we replace interests now with go to dev card but still need a value set for interests?

pages/start.tsx
```diff
@@ -70,7 +68,7 @@ const LoginStep1: React.FC<LoginStep1Props> = ({ user }) => {
     } else if (onboarded === false && user && providerToken) {
       setQueryParams({ step: "2" } satisfies QueryParams);
     }
-  }, [user, onboarded, providerToken]);
+  }, [user, onboarded, providerToken, router]);
```
nickytonline, line 71: did this get added because of a linter warning?

nickytonline (approved):
Works great on mobile and desktop. 🚢

pages/start.tsx
brandonroberts: Yes, I can take it out though. I thought it was an error

pages/start.tsx
brandonroberts: Yes, its still required in the API for now

pages/start.tsx
nickytonline: It's because it's used in the useEffect. Was just checking is all. Feel free to keep it in.

pages/start.tsx
nickytonline: I created https://github.com/open-sauced/api/issues/1030 for this.

isabensusan (approved):
I'd say it's good to ship as is! 

But if possible, is there any chance we can add a preview of the card in this step? Even the OG image could work. Or as a second resource, a random example image?

(very rough mock!! but you get the idea)

image

BekahHW (changes requested):
My concern here is that we don't have a clear path for what they should do after they share their devcard and we give them no other options in the onboarding path. Can we give an option? Share your DevCard or Get Started Now! The latter links to the Workspace. 

It would also be nice if once a person shares a DevCard during onboarding, they're taken to the profile edit page. 

We don't want to leave them hanging after they share. 

I'm not trying to nitpick, but I want to make sure users can see a clear path to usage.

brandonroberts:
> My concern here is that we don't have a clear path for what they should do after they share their devcard and we give them no other options in the onboarding path. Can we give an option? Share your DevCard or Get Started Now! The latter links to the Workspace.
> 
> It would also be nice if once a person shares a DevCard during onboarding, they're taken to the profile edit page.
> 
> We don't want to leave them hanging after they share.
> 
> I'm not trying to nitpick, but I want to make sure users can see a clear path to usage.

Good point about the workspace destination. I think having a second button that goes to the workspaces page is good, it has a modal that pops up introducing workspaces in a video

bdougie:
> Good point about the workspace destination. I think having a second button that goes to the workspaces page is good, it has a modal that pops up introducing workspaces in a video

Second option should view your profile. If you logged in with tour GitHub, chances are you will want to see your profile.

Workspaces is a stretch today, and the scope for making that work from onboarding is something we should address just not in this PR

brandonroberts:
Updated the last step to preview your DevCard, and added a button to go to your profile

image

BekahHW (approved):

PR #4010, opened 2024-08-22
fix: Prevent OSCR score from popping out of dev card with long usernames

### LINKED ISSUE #3981: Bug: Dev Card with long username cause OSCR to pop out of card
A dev card with long username causes the OSCR score to pop out of card. You can see an example of this on Josh's dev card page. https://app.opensauced.pizza/u/JoshuaKGoldberg/card

CleanShot 2024-08-18 at 13 34 15

Note that the OG image is fine, as we don't add the OSCR score beside the username.

Mandeep56Singh:
## Description

This PR addresses the bug where a long username in a developer card causes the OSCR score to overflow and pop out of the card's boundary while also address related to it.

- resolve Hydaration error in Devcard component by using useEffect and useState for props.isLoading.
- fix styling in front view of card by limiting its width,text-center and wrapping
- fix styling in back view of card by truncate tailwind utility
- increase font size of oscr  limit  from 0.5 rem to 0.7rem to make it visible

closes #3981 

### Front view  at max username length
Screenshot (438)

### Back view at lengthy username
Screenshot (435)

### username in next line 
Screenshot (437)

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [ ] Tier 4

### DIFF
--- a/components/molecules/DevCard/dev-card.tsx
+++ b/components/molecules/DevCard/dev-card.tsx
@@ -160,7 +160,9 @@ export default function DevCard(props: DevCardProps) {
                 height={100}
                 className="rounded-full border-[1.5px] border-gray-300 border-opacity-40"
               />
-              <p>@{props.devstats?.login}</p>
+              <p className="break-words mx-auto whitespace-normal text-center max-w-[200px]">
+                @{props.devstats?.login}
+              </p>
             </div>
 
             {/** OSCR Score **/}
@@ -194,10 +196,10 @@ export default function DevCard(props: DevCardProps) {
                 height={50}
                 className="absolute rounded-full border-[1.5px] border-gray-300 border-opacity-40 -left-1"
               />
-              <p className="pl-10">{props.devstats?.login}</p>
+              <p className="pl-10 truncate items-center">{props.devstats?.login}</p>
               <p className="-pl-4 justfy-self-end text-sm flex gap-0.5">
                 {Math.ceil(props.devstats?.oscr || 0)}
-                <span className="text-[0.5rem] text-gray-100 font-normal">/300</span>
+                <span className="text-[0.7rem] text-gray-100 font-normal">/300</span>
               </p>
             </div>
 
--- a/components/molecules/DevCard/dev-card.tsx
+++ b/components/molecules/DevCard/dev-card.tsx
@@ -191,7 +191,7 @@ export default function DevCard(props: DevCardProps) {
                 className="absolute rounded-full border-[1.5px] border-gray-300 border-opacity-40 -left-1"
               />
               <p className="pl-10 truncate items-center">{props.devstats?.login}</p>
-              <p className="-pl-4 justfy-self-end text-sm flex gap-0.5">
+              <p className="-pl-4 justify-self-end items-center text-sm flex gap-0.5">
                 {Math.ceil(props.devstats?.oscr || 0)}
                 <span className="text-[0.7rem] text-gray-100 font-normal">/300</span>
               </p>
--- a/components/molecules/DevCard/dev-card.tsx
+++ b/components/molecules/DevCard/dev-card.tsx
@@ -28,7 +28,7 @@ export type DevCardProps = {
 export default function DevCard(props: DevCardProps) {
   const [isFlipped, setIsFlipped] = useState(props.isFlipped ?? false);
   const isInteractive = props.isInteractive ?? true;
-  const isLoaded = useHasMounted();
+  const hasMounted = useHasMounted();
 
   useEffect(() => {
     setIsFlipped(props.isFlipped ?? false);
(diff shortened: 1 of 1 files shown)


nickytonline:
I was going to suggest not showing the OSCR on the back of the card since it's already on the front and leave as much real estate for the username. Thoughts @brandonroberts @zeucapua @isabensusan?

Mandeep56Singh:
@brandonroberts  actually I had hydration error while solving this issue, To my understanding It was due to props.isLoading. 
However I have fixed it. I would require your assistance on this.

brandonroberts:
> I was going to suggest not showing the OSCR on the back of the card since it's already on the front and leave as much real estate for the username. Thoughts @brandonroberts @zeucapua @isabensusan?

I agree

BekahHW:
@nickytonline we have it on the back of the card like that to provide more context - it's out of 300. I think if we remove it here we need to have it referenced somewhere else on the card.

nickytonline:
> @nickytonline we have it on the back of the card like that to provide more context - it's out of 300. I think if we remove it here we need to have it referenced somewhere else on the card.

Maybe we move it somewhere else on the card on the back. It looks nice to see a full username on the back instead of it being cut off.

Mandeep56Singh:
@BekahHW we can it in front, we just need to add a line displaying 
there are various ways for it 
 
 ### we can display the out of message under score, but when we have long username ( 2 lines ) then it will look ugly
Screenshot (442)

### displaying it with OSCR score
Screenshot (440)

### I found this good, since we don't really need "score" in text as OSCR itself is rating 
Screenshot (443)

Mandeep56Singh:
@nickytonline  full username on the back can take a lot of space as longest username can be of 39 characters, it would require two lines. I think  adding  more statistics on the back side, so user can know more about the person. We can remove the username, profile pic and oscr score from back. 

we have different option for it.
###  added new statistics and remove header of back view
Screenshot (444)

### this looks good
Screenshot (445)

what are your thoughts ?

BekahHW:
I would prefer the original fix here. It solves the initial problem without a redesign. Maybe @isabensusan has thoughts.

isabensusan:
@Mandeep56Singh I also prefer the original original fix! Let's truncate longer user names

Mandeep56Singh:
@isabensusan  ok, the front view also have overflowing problem, the longest github username can be 39 characters long, so it will be two lines, I thought of centering the username. 
what you are your thoughts ?
thanks.

Screenshot (437)
Screenshot (438)

isabensusan:
hey @Mandeep56Singh ! Yes, centering looks good  :)

Mandeep56Singh:
> hey @Mandeep56Singh ! Yes, centering looks good :)

Thanks a lot, @isabensusan! I really appreciate the feedback. 
Looking forward to making more contributions to OpenSauced!"

isabensusan (approved):
Design wise, looks good to me!

components/molecules/DevCard/dev-card.tsx
```diff
@@ -38,6 +39,13 @@ export default function DevCard(props: DevCardProps) {
     }
   }, [props.isInteractive, props.isFlipped]);
 
+  // used to solve hydartion error
+  useEffect(() => {
```
brandonroberts, line 43: We have a hook for this already named `useHasMounted` that can be used instead.

```ts
import { useHasMounted } from "lib/hooks/useHasMounted";

const hasMounted = useHasMounted();
```

brandonroberts (dismissed):

components/molecules/DevCard/dev-card.tsx
```diff
@@ -186,10 +190,10 @@ export default function DevCard(props: DevCardProps) {
                 height={50}
                 className="absolute rounded-full border-[1.5px] border-gray-300 border-opacity-40 -left-1"
               />
-              <p className="pl-10">{props.devstats?.login}</p>
+              <p className="pl-10 truncate items-center">{props.devstats?.login}</p>
               <p className="-pl-4 justfy-self-end text-sm flex gap-0.5">
```
nickytonline, line 194: Center the text. Also it looked like it was never self justifying to the end.

```suggestion
              <p className="-pl-4 justify-self-end items-center text-sm flex gap-0.5">
```

nickytonline (changes requested):
Works great @Mandeep56Singh! One small change request and this is good to go.

CleanShot 2024-09-18 at 08 41 44

nickytonline (approved):
Thanks for making the changes @Mandeep56Singh! 🚢

brandonroberts (approved):

nickytonline:
When you have a chance @zeucapua, we need your approval. Thanks!

PR #4018, opened 2024-08-23
fix: `RossChart` order changing on filter

### LINKED ISSUE #3993: Bug: Contributors chart order changes on filter
### Describe the bug

When using the contributors filter, the data gets recalculated via `useMemo` and seems to 'sometimes' change the order of the time buckets.

https://github.com/user-attachments/assets/c2bc652b-16c2-4b70-9a9e-88aee0f5cff7

### Steps to reproduce

1. Go to any repository page (preferably one with multiple buckets)
2. Click through the filters on the Contributors chart
3. See the dates changing order between filters

zeucapua:
## Description

Uses `toReversed()` function to create a reversed copy of `stats.contributors` instead of using `reversed()` in place, which causes the data to flip on every filter change

Closes #3993

**Before**

https://github.com/user-attachments/assets/c2bc652b-16c2-4b70-9a9e-88aee0f5cff7

**After**

https://github.com/user-attachments/assets/f9c5ce67-a018-430e-b3b4-e6fbde6fb7da

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/components/Repositories/RossChart.tsx
+++ b/components/Repositories/RossChart.tsx
@@ -45,28 +45,26 @@ export default function RossChart({ stats, isLoading, error, range, onFilterClic
   }, [stats]);
 
   const weeklyData = useMemo(() => {
-    // @ts-ignore
-    const contributors = stats?.contributors.toReversed() ?? [];
-
-    // @ts-ignore
-    return contributors.map((week) => {
-      return {
-        new: filterOutside ? week.new : 0,
-        recurring: filterRecurring ? week.recurring : 0,
-        internal: filterInternal ? week.internal : 0,
-        bucket: new Date(week.bucket).toLocaleDateString(undefined, { month: "numeric", day: "numeric" }),
-      };
-    });
+    return stats?.contributors
+      .slice()
+      .reverse()
+      .map((week) => {
+        return {
+          new: filterOutside ? week.new : 0,
+          recurring: filterRecurring ? week.recurring : 0,
+          internal: filterInternal ? week.internal : 0,
+          bucket: new Date(week.bucket).toLocaleDateString(undefined, { month: "numeric", day: "numeric" }),
+        };
+      });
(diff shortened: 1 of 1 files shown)


components/Repositories/RossChart.tsx
```diff
@@ -45,7 +45,11 @@ export default function RossChart({ stats, isLoading, error, range, onFilterClic
   }, [stats]);
 
   const weeklyData = useMemo(() => {
-    return stats?.contributors.reverse().map((week) => {
+    // @ts-ignore
```
zeucapua, line 48: `Array.toReversed()` has been supported by all major browsers since last year; I don't know why Typescript/tsserver do not recognize it. Adding `@ts-ignore` to remove build errors

jpmcb: > I don't know why Typescript/tsserver do not recognize it

I wonder if we have an older version of those types / tools that don't yet have full GA support for it yet. Regardless, this is fine 👍🏼 I'd maybe add a small comment for future us to remove that `ts-ignore` when we consume a version of the tooling that supports this

jpmcb (approved):

brandonroberts (approved):

components/Repositories/RossChart.tsx
jpmcb: `ts-ignore` may also just be taking the conservative approach: while it's been supported for well over a year, not everyone upgrades their browsers that often. Still, I'd say this is fine!

brandonroberts: Would using `slice` before doing the `reverse` help?  As `reverse` mutates the array in place, where `slice` would create a new copy first.

zeucapua: Yeah if we want to remove the `@ts-ignore`, `slice` then reverse would do it. I'll get a quick commit in for it

PR #4043, opened 2024-08-27
fix: now if loading user profile fails an error is thrown instead of 404ing

### LINKED ISSUE #4042: Bug: User profile 404ing when in fact it was a server error
This is just a hypothesis, but my user profile in prod was 404ing for about a good 5-6 minutes and finally displayed.

The current code looks like this in getServerSideProps for the user profile page. We should check if the request to the API succeeded (even if a 404) or if it failed.

```diff
export const getServerSideProps = async (context: UserSSRPropsContext) => {
  const { username } = context.params ?? { username: "" };

  if (!isValidUrlSlug(username)) {
    return { notFound: true };
  }

  const req = await fetch(`${process.env.NEXT_PUBLIC_API_URL}/users/${username}`, {
    headers: {
      accept: "application/json",
    },
  });

  if (!req.ok) {
-    return { notFound: true };
+    if (req.status === 404) {
+      return {
+        notFound: true,
+      };
+    }
+
+    const errorText = `Failed to load user profile: ${username}`;
+    captureException(new Error(errorText, { cause: req.statusText }));
+    throw new Error(errorText);
  }

  const userData = (await req.json()) as DbUser;
  const ogImage = siteUrl(`og-images/dev-card`, { username });

  // Cache page for 60 seconds
  context.res.setHeader("Cache-Control", "public, max-age=0, must-revalidate");
  context.res.setHeader("Netlify-CDN-Cache-Control", "public, max-age=0, stale-while-revalidate=60");
  context.res.setHeader("Cache-Tag", `user-profiles,user-profile-${username}`);

  return {
    props: { username, user: userData, ogImage },
  };
};
```

nickytonline:
## Description

Now if loading user data fails on a user profile page, an error is thrown instead of 404ing as it's misleading if an existing user is said to not be found.

Fixes #4042

## Steps to QA

Kind of hard to test this atm. The API can 404 if a user isn't found, but then it goes to the stub a user and if that fails it gives a 400.

1. Go to a valid user profile, e.g. /u/brandonroberts
2. Page loads
3. Go to an invalid user profile, e.g. /u/brandonroberts____
4. It doesn't 404, it throws a 400

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/pages/u/[username]/index.tsx
+++ b/pages/u/[username]/index.tsx
@@ -4,6 +4,7 @@ import { jsonLdScriptProps } from "react-schemaorg";
 import { Person } from "schema-dts";
 
 import { useRouter } from "next/router";
+import { captureException } from "@sentry/nextjs";
 import SEO from "layouts/SEO/SEO";
 
 import useContributorPullRequests from "lib/hooks/api/useContributorPullRequests";
@@ -118,7 +119,15 @@ export const getServerSideProps = async (context: UserSSRPropsContext) => {
   });
 
   if (!req.ok) {
-    return { notFound: true };
+    if (req.status === 404) {
+      return {
+        notFound: true,
+      };
+    }
+
+    const errorText = `Failed to load user profile: ${username}`;
+    captureException(new Error(errorText, { cause: req.statusText }));
+    throw new Error(errorText);
   }
 
   const userData = (await req.json()) as DbUser;


pages/u/[username]/index.tsx
```diff
@@ -118,7 +119,15 @@ export const getServerSideProps = async (context: UserSSRPropsContext) => {
   });
 
   if (!req.ok) {
-    return { notFound: true };
+    if (req.status === 404) {
```
Comment (nickytonline) [inline pages/u/[username]/index.tsx:122]: This logic makes sense in the page @brandonroberts @jpmcb, but the reality is if the API is bogged down, it fails to get the user info for a valid user and will error (good), but if it's an invalid user, we 400 instead of 404. TLDR; we'll never get a 404 if `!req.ok`. The check up further for a valid slug will 404 if it's not a valid github username though.

TLDR; this is a good fix if a valid user profile fails to load, but for a non existent user with a valid GitHub username format will also throw.

brandonroberts (approved):

pages/u/[username]/index.tsx
Comment (jpmcb) [reply pages/u/[username]/index.tsx:122]: This sounds like the API needs to handle its errors better instead of the App trying to conform to the API's poor error handling. Taking a quick look at the code there, if the GitHub client code fails during user stubbing (for any reason like a 500, 400, etc.), the API will just throw a bad request exception.

This could probably have a pattern match on the GitHub request status code and bubble that up from there (i.e., if GitHub 404s on a user not being there, we thrown that not found exception).

Comment (nickytonline) [reply pages/u/[username]/index.tsx:122]: Yeah definitely some changes to the API are in order, but on the page level, I think the logic here makes sense.

jpmcb (approved):

PR #4062, opened 2024-08-28
feat: use `SplitButton` for repo page Add Workspace buttons

### LINKED ISSUE #3942: Feature: Use Split Button Component for SBOM and Repository Workspace Creation on Repository Page
> @bdougie I was going to say the same thing, two primary buttons is a bit too much. I'm ok with the secondary dark version to get it out the door, but the follow up should definitely add a dropdown button.
> 
> However we do have the [dropdown component](https://design-insights.netlify.app/?path=/story/design-system-atoms-singleselect--default), so I wonder if this would be just good enough for now? For example: image
> 
> And we can work on a similar component but with a primary button + a down arrow as the dropdown trigger for a follow up

Yeah I think that's the SingleSelect component? I can do a fast follow up on this using that.

_Originally posted by @nickytonline in https://github.com/open-sauced/app/issues/3938#issuecomment-2286872740_
            
Depends on #4041 (should be merged soon)

zeucapua:
## Description

Implements the `SplitButton` component to combine the 'Add to Workspace' and 'Create Workspace from SBOM' buttons on desktop. Adds a new optional 'icon' prop to `SplitButton`. 

Note: this is only for the desktop view. On mobile we are using the `Drawer` component which requires its own trigger button, so it can't be separated to an external trigger.

Closes #3942 

https://github.com/user-attachments/assets/2db7c732-d72a-48de-8eb7-76414b81734e

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/pages/s/[org]/[repo]/index.tsx
+++ b/pages/s/[org]/[repo]/index.tsx
@@ -43,10 +43,10 @@ import YoloChart from "components/Repositories/YoloChart";
 import LanguagePill, { getLanguageTopic } from "components/shared/LanguagePill/LanguagePill";
 import OssfChart from "components/Repositories/OssfChart";
 import LotteryFactorChart from "components/Repositories/LotteryFactorChart";
-import Tooltip from "components/atoms/Tooltip/tooltip";
 import { CopyContainer } from "components/shared/CopyContainer";
 import CopyImageBranding from "components/shared/CopyImageBranding";
 import { getRepositoryOgImage, RepositoryOgImage } from "components/Repositories/RepositoryOgImage";
+import { SplitButton } from "components/shared/SplitButton";
 
 const AddToWorkspaceModal = dynamic(() => import("components/Repositories/AddToWorkspaceModal"), {
   ssr: false,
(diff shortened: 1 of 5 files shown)


pages/s/[org]/[repo]/index.tsx
```diff
@@ -248,36 +248,22 @@ export default function RepoPage({ repoData, ogImageUrl }: RepoPageProps) {
...
-                            >
-                              Learn more...<span className="sr-only"> about SBOM</span>
-                            </a>
-                          </div>
-                        }
-                      >
-                        <Button
-                          variant="dark"
-                          onClick={() => {
+                    <SplitButton
```
Comment (nickytonline) [inline pages/s/[org]/[repo]/index.tsx:251]: The split button works, but it looks like now that it's being used, it causes some layout shift.

CleanShot 2024-08-28 at 18 05 54

nickytonline (changes requested):
Not specific to this PR but since it's the first usage of the split button component, the shifting should get sorted out either in a separate PR or this one.

zeucapua:
It shouldn't shift now; aligned the dropdown menu to "end", bringing it more towards the center of the page

image

brandonroberts (approved):

nickytonline (changes requested):
It's line up now component wise, but there is still some side shifting. Happy to pair if you want. We have our sync now anyway.

CleanShot 2024-08-29 at 14 51 55

zeucapua:
Looks like under the hood, radix uses the `react-remove-scroll` package to remove scrolling when the Dropdown/SingleSelect is opened. The fix to keep the scroll is to set the prop `modal={false}` on the root components.

nickytonline (approved):
Works great! Thanks @zeucapua! 🚢 

CleanShot 2024-08-30 at 16 29 13

PR #4091, opened 2024-09-10
chore: 2.62.0 release

nickytonline:


zeucapua (approved):

nickytonline:
Hmm, even after the artifact action updates to v4, the app release is still giving the same error. I can cut another PR, but I feel like it'll result in the same thing. I've restarted the job a few times, so no dice there. I might just do a new PR anyway.

PR #4158, opened 2024-10-10
fix: use a tags instead of link tags

### LINKED ISSUE #4108: Bug: loading dev card page from the user profile crashes
Going to the dev card page directly works, but if you click on the dev card button near the avatar on the user profile page, it crashes.

CleanShot 2024-09-18 at 08 17 58

CleanShot 2024-09-18 at 08 20 19

The `user` is `null` causing the crash.

Bashamega:
## Description

Fixed the bug by using a tags instead of link tags

Fixes #4108 

https://github.com/user-attachments/assets/2310dc7d-f281-4dd5-ad5f-3629c0309d6b

### DIFF
--- a/components/molecules/ContributorProfileHeader/contributor-profile-header.tsx
+++ b/components/molecules/ContributorProfileHeader/contributor-profile-header.tsx
@@ -156,14 +156,14 @@ const ContributorProfileHeader = ({
           />
 
           <Tooltip content="Get dev card">
-            <Link
+            <a
               href={cardPageUrl(username!)}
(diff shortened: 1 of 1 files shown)


components/molecules/ContributorProfileHeader/contributor-profile-header.tsx
```diff
@@ -156,14 +156,14 @@ const ContributorProfileHeader = ({
           />
 
           <Tooltip content="Get dev card">
-            <Link
+            <a
```
nickytonline, line 159: Although we lose the SPA like mode by going to an anchor, this is better and it's only one page navigation.

nickytonline (approved):
Thanks @Bashamega!

